Issue #26, opened 2022-05-10
Refactor chunks to S4


gogonzo:
Current chunks version generates multiple problems. The most important is that: 
- Chunks is an R6 object which is not mutable. This means that changes in the chunks don't trigger shiny reactivity. To achieve this we need to clone or force some reactive events using reactive fields [kept in chunks](https://github.com/insightsengineering/teal.code/blob/f7d0cdd870d06b359883c17aec263293a8fdd034/R/chunks.R#L171)

- Partial consequence of above is that chunks having queue system which delays evaluation of the code. This causes the mismatch between code evaluation and reactivity and creates general confusion.

- Problem with chunks object in the session is a minor and can be solved in the easy yet time consuming way https://github.com/insightsengineering/teal.code/issues/22

Solution is to refactor chunks to S4 object which will be "reactive" by default - changing any slots in the S4 object changes the address of the S4 object.

## class code 
```r
setClass(
  "Quosure",
  representation(expr = "expression", env = "environment"),
  prototype(expr = expression(), env = new.env(parent = parent.env(.GlobalEnv)))
)
setGeneric("push", function(object, expr, name = "code") {
  standardGeneric("push")
})
setGeneric("get_var", function(object, var) {
  standardGeneric("get_var")
})
setGeneric("get_code", function(object, deparse = FALSE) {
  standardGeneric("get_code")
})
setGeneric("join", function(object, object2) {
  standardGeneric("join")
})
setMethod("push", signature("Quosure"), function(object, expr, name) {
  # combine expressions
  nm <- make.unique(c(names(object@expr), name))
  object@expr <- setNames(c(object@expr, expr), nm)

  # need to copy the objects from old env to new env
  # to avoind "chunks" sharing the same environment
  env <- new.env()
  for (i in ls(object@env)) {
    env[[i]] <- object@env[[i]]
  }
  eval(expr, envir = env)
  object@env <- env
  object
})
setMethod("get_var", signature("Quosure", "character"), function(object, var) {
  get(var, envir = object@env)
})
setMethod("get_code", signature("Quosure"), function(object, deparse) {
  if (deparse) {
    deparse(object@expr)
  } else {
    object@expr
  }
})
setMethod("join", signature("Quosure", "Quosure"), function(object, object2) {
  # object2 can't have modified object of the same name! See chunks_push_chunks
  common_names <- intersect(ls(object@env), ls(object2@env))
  is_identical_obj <- vapply(common_names, function(x) {
[...]

PR #49, opened 2022-07-06
Chunks replacement

gogonzo:
refers to #26 
Lighweight code/env method to evaluate code in. Not important during this sprint - I'm sharing just to discuss stuff
New class Quosure having `code` and `env` slots and following methods:
1. `eval_code` to `eval` and append the `code`
2. `get_var` and `[[` methods to extract "var" from the environment
3. `get_code` to obtain the code.
4. `join` to join two Quosures see what we [use to do in chunks](https://github.com/insightsengineering/teal.code/blob/923e281ebcde875d1c043f7a3669349af8a0bba2/R/chunks.R#L211) This is a key point in this class, of how to validate that `env` and `code` can be joined together.


R/Quosure.R
```diff
@@ -0,0 +1,270 @@
+#' Reproducible class with environment and code.
+#'
+#' @slot code (`character`) to reproduce the environment
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#' of the `code` slot.
+setClass(
+  "Quosure",
+  representation(code = "character", env = "environment"),
+  prototype(code = character(0), env = new.env(parent = parent.env(.GlobalEnv)))
```
pawelru, line 9: I would suggest to use expression / call type instead which is much more closer to the _code_ representation in R.

gogonzo: We should rather handle code as character as it supports comments, empty lines etc. `language` classes discards comments information.

pawelru: The question is: do we really need to have a comments, empty lines etc? The value is very questionable for me. I feel that it is a _code_ that we need to capture, evaluate and return - not a _script_ that produces is. I also think that it would be better to stick to the more raw and simple structures as managing characters might be unpredictable and could lead into some unforeseen issues such as char encoding, empty lines etc. Character should be _just_ a representation of code at the very end and such casting should be done by shiny module - not the class.

R/Quosure.R
```diff
@@ -0,0 +1,270 @@
+#' Reproducible class with environment and code.
+#'
+#' @slot code (`character`) to reproduce the environment
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#' of the `code` slot.
+setClass(
+  "Quosure",
```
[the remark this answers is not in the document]
nikolas-burkoff: I think tealEnv isn't bad or you could go further and call it a tealReactive

R/Quosure.R
```diff
@@ -0,0 +1,270 @@
...
+#'  should overwrite their equivalents in the environment of the `object`.
+#' @export
+setGeneric("join", function(object, object2, overwrite = FALSE) {
+  standardGeneric("join")
+})
+
+setMethod("join", signature("Quosure", "Quosure"), function(object, object2, overwrite = FALSE) {
+  # todo: revise chunks_merge_chunks and make robust join method
+
+  # object2 can't have modified object of the same name! See chunks_push_chunks
```
nikolas-burkoff, line 194: could this be too restrictive?

R/Quosure.R
```diff
@@ -0,0 +1,270 @@
...
+
+#' Copy environment content to the other environment
+#' To avoid sharing the same environment by multiplequo object (in reactives)
+#' @keywords internal
+.copy_env <- function(env) {
+  # need to copy the objects from old env to new env
+  # to avoid "chunks" sharing the same environment
+  env_new <- new.env()
+  for (i in ls(env)) {
+    env_new[[i]] <- env[[i]]
```
nikolas-burkoff, line 232: - Could this be really slow? What happens if there's lots of data here?  Could you handle this by having the parent of the new environment be say the old ones? Though I guess that could get more complex. I could also imagine other complex things that could but probably shouldn't be done here. 

- Do you actually want separate environments for everything - could you have a separate single "evaluation environment" for the shiny module? Would that be easier for module developers to grasp or harder?

- Also what happens if the parent environments of the two objects don't match - you could get some weird behaviour joining them in that case? 

I guess I'm trying to get as intuitive (if you already know shiny) behaviour as possible here - could the output of the quosure actually be available in the module's environment? So that they behave as similar to reactives as possible.

R/Quosure.R
```diff
@@ -0,0 +1,270 @@
...
+
+  # if  B has extra code which is not  at the end
+  # only one object can have extra code at the
+
+  # if B has extra code at the end and no object conflict
+  # if A has extra code at the end and no object conflict
+  #
+
+  code_only_in_a <- which(!code %in% code2)
+  code_only_in_b <- which(!code2 %in% code)
```
nikolas-burkoff, line 252: These feels risky - how about?

```
for ( i in 1:10) {
  x[[i]] <- 5
}
```

and

```
for ( i in 1:10) {
  print("Hello gogonzo")
}
```

It feels like we should be able to leverage shiny's reactivity handling here as that doesn't seem to have a problem knowing what depends on what??

nikolas-burkoff (commented):
A few musings from me, some may be useful, others less so

R/Quosure.R
gogonzo: Are you aware that current implementation which returns SRC with comments and new-line-spaces will not be possible anymore?
What in your opinion would be unpredictable?:
- `"var <- 1"` - code as character
- `"# this is a comment"` - code (comment) as character
- `""` - empty line - code (empty) as character

We will support primarily:
- `var <- 1` - unquoted expression (to be substituted)
- `substitute(var <- 1, list(var = quote(new_var))` - quoted expression
- `rlang::expr(!!new_var <- 1)` - quoted expression

To possibly add a comment one would have to use `parse` which will return an empty `expression()` but we can get parsed text from the attr(expression, "wholeSrcref")
```r
parse(text = "
# comment

# new comment
")

#> expression()
```



Accepting only expression doesn't prevent anyone from executing `"\n"` which is a valid R expression and will add an artifact to the code (it won't be extra line but "\n" visible as is in the code.
```r
eval_code(Quosure, quote("\n"))
```

pawelru: > Are you aware that current implementation which returns SRC with comments and new-line-spaces will not be possible anymore?

Yes I am and this is what I put in question - I am not so sure if it brings a lot of value to the users but it changes drastically the implementation. The problems with char management differs significantly from problems of code management and I feel that code management would be easier, simpler and closer to the actual task.
An example of unexpected issues is regular expressions applied on code as char which doesn't seem to be an appropriate way of managing the code. If we store code as char then what is stopping devs to manipulate the data with regular expression? Well... nothing! Same for creating inputs for this class. I want module developers to create code/expressions - not chars. Expressions are safer and char can accept anything. See below:
```
> expression(1a1)
Error: unexpected symbol in "expression(1a1"
> "1a1"
[1] "1a1"
```

> To possibly add a comment one would have to use ...

How about no comments at all? At the end of the day, SRC returns a _code_ not a _script_. I know I representing some extreme view right now (and I am feeling like a devil advocate) but I need to ask some basic questions because we might just blindly assume something that is not true (anymore).

gogonzo: There is some way to extract the `srcref` from the parsed expression to have the comments or the line-breaks so it can be preserved (and combined) in the expr attribute. This is diffucult task but possible In my opinion to achieve. Only adding comments and new line would have to be done through `eval_code(parse(text = "any code"))`.

```r
parse(text = "
  #

  elo <- 1
") |> attr("wholeSrcref")

#> #
#>
#> elo <- 1
```

R/Quosure.R
```diff
@@ -0,0 +1,307 @@
...
+#'
+#' Given code is evaluated in the `Quosure` environment and appended to the `code` slot. This means
+#' that state of the environment is always a result of the stored code (if `Quosure` was initialized)
+#' with reproducible code.
+#'
+#' @name eval_code
+#'
+#' @param object (`Quosure`)
+#' @param code (`character(1)` or `language`) code to evaluate. Also acceptps and stores comments
+#' @param name (`character(1)`) name of the code element.
```
pawelru, line 92: what's the point of this argument?

R/Quosure.R
```diff
@@ -0,0 +1,307 @@
...
+  standardGeneric("get_code")
+})
+
+setGeneric("join", function(object, object2, overwrite = FALSE) {
+  standardGeneric("join")
+})
+
+#' Evaluate the code in the `Quosure` environment
+#'
+#' Given code is evaluated in the `Quosure` environment and appended to the `code` slot. This means
```
pawelru, line 84: > and appended to the `code` slot

What if we want to store different code that the one evaluated? I am thinking about obfuscation of passwords etc.

R/Quosure.R
gogonzo: to name the elements of the code (to name a vector). Might be needed to do something like `get_eval_details_srv ` . It will be possible also to keep more informations like warnings, errors and notes also from the evaluations

R/Quosure.R
gogonzo: I don't have an idea how to enable this. I think in DDL we should rather do:
```r
expr <- rice_connect(login = login_val, password = password_val)
expr_to_eval <- substitute(expr, list(login = input$login, password = input$password))
expr_to_return <- substitute(expr, list(login = Sys.getvar("USER"), password = quote(askpass::askpass()))
new_quosure(eval(expr), code = expr_to_return))
```

Don't you think that expr replacement is little to much to the Quosure?

R/Quosure.R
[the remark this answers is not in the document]
pawelru: I am thinking about the name and the key question is that if it is code + env or rather env + code? I think it's the latter one as it conceptually _just_ adds a code attribute to the environment. Therefore:
- envcode
- envc ("c" as a shortcut of "code")
- reprenv - this I don't like that much because the code might not necessary reproduce the environment (vide: password obfuscation; also code needs a (separate) environment on which it is evaluated - i.e. rlang' quosure)

(I am going to modify this post with new ideas)

R/Quosure.R
pawelru: Do we have to name elements of the code vector? Maybe YAGNI and start simple?

R/Quosure.R
pawelru: I like your idea and I agree Quosure shouldn't be responsible for expr replacement. It needs to be done on client side (inside the module) or... a separate class that Quosure could use.

```
expr <- rice_connect(...)
expr_obj <- obfuscate_substitute(expr, list(path = input$path, login = input$login, password = input$password), list(login = ..., password = ...)
new_quosure(eval(expr_obj$code), expr_obj$obfuscated_code)
# or alternatively
new_quosure(expr_obj)
```

All in all, quosure API needs to allow passing those two values - either combined into the single object or provided separately.

R/Quosure.R
```diff
@@ -0,0 +1,307 @@
...
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(code = "character", env = "environment"),
+  function(code = character(0), env = parent.env(.GlobalEnv)) {
+    checkmate::check_class(env, "environment")
+    checkmate::check_string(code)
```
pawelru, line 42: I would say that these checks are obsolete as the arg checks are done by method dispatch

UPDATE:
maybe leave string check because it includes length validation

R/Quosure.R
```diff
@@ -0,0 +1,307 @@
...
+#' reproducibility
+#' - more cases todo
+#' @param object (`Quosure`)
+#' @param object2 (`Quosure`)
+#' @export
+setGeneric("join", function(object, object2, overwrite = FALSE) {
+  standardGeneric("join")
+})
+
+setMethod("join", signature("Quosure", "Quosure"), function(object, object2) {
```
pawelru, line 200: I can see that you do more of a _combine_ in which you treat both quosures the same. Do we need a _join_ instead? As a _join_ I mean merge one into the other (vide: joining the tables - right table into the left one). _Join_ would be much more simpler (i.e. `parent.env(right) <- left`) and you don't have to track all the edge cases.

Conceptually, what would be the output of the following (pseudo-code):
```
q1 <- new_quosure("x <- 1", new.env()$x <- 1)
q2 <- new_quosure("x <- 2", new.env()$x <- 2) 
q <- join(q1, q2)
```

You have hard stop. Don't we want `q@env$x == 2; q@code == "x<-1; x<-2"` instead?

Please think of the class role. It needs to track the environment together with the code generating it. The above scenario is perfectly fine workflow thus I would expect it not to fail.

R/Quosure.R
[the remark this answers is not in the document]
nikolas-burkoff: > tealReactive - as above + "reactive" indicates some shiny relationship which is not true

wishful thinking as a user of shiny I would love to be able to treat these objects in as similar way as possible to shiny reactives would make it super easier for people with shiny experience to use them

R/Quosure.R
gogonzo: I think you are right. If we consider two quosures to be reproducible (code ~ env). Then _q1 ∪ q2_ is a _code1 ∪ code2_ and _env1 ∪ env2_ (on the overlap object from env2 overwrites object from env1).

gogonzo: I don't understand the comment with combine, collapse, join. Do you think join is a wrong name?

pawelru: > I don't understand the comment with combine, collapse, join. Do you think join is a wrong name?

For _join_ the argument order matters as you put the _right_ one into the _left_ one - therefore the _right_ one has higher importance that the _left_ one as it might overwrite it.
For combine/collapse you tend to treat both _left_ and _right_ the same way and try to raise an error on conflicts.

Name should reflect the behaviour. Looking at the current function body it's more a combine/collapse than join. But I feel we agree to have join behaviour instead therefore the name could stay as it is now.

R/Quosure.R
gogonzo: > Could this be really slow? What happens if there's lots of data here? Could you handle this by having the parent of the new environment be say the old ones? Though I guess that could get more complex. I could also imagine other complex things that could but probably shouldn't be done here.

It's not slow. It's making a new reference in the new env to the same pointer. 

> I guess I'm trying to get as intuitive (if you already know shiny) behaviour as possible here - could the output of the quosure actually be available in the module's environment? So that they behave as similar to reactives as possible.

Do you mean that objects from the quosure@env would be automatically or not available in the reactive?
We can still use `q[["object"]]` instead. I don't think mixing these two environments is a good idea.
1. Code could be successful if some object was in the reactive environment and not in quosure@env. 
2. We will need a way to pass `quosure@env` to the `reactive(..., env = )`
3. We don't want to include objects in the quosure@env which are not generated by the code. Often in the reactives there are objects which are not part of the code (proxy variables, observers)

> Do you actually want separate environments for everything - could you have a separate single "evaluation environment" for the shiny module? Would that be easier for module developers to grasp or harder?
I don't get it. Let's talk about this on thursday

R/Quosure.R
gogonzo: I think order matters as `y` can overwrite `x`. The most appropriate name in my opinion for them is `union` but probably we shouldn't register a S4 method union as there is already S3. I think `join` also makes sense because we don't combine only but also resolve conflicts between duplicated code and objects.

R/Quosure.R
```diff
@@ -0,0 +1,285 @@
...
+  # - remove duplicated code from object2 and append to the object1
+  unique_expr <- object2@code[!(object2@code %in% object@code & names(object2@code) %in% names(object@code))]
+  # - in case object@code and object2@code contains extra code with the same name - code2 needs to be renamed
+  object@code <- .keep_code_name_unique(object@code, unique_expr)
+
+  # insert (and overwrite) objects from env2 to env
+  object@env <- .copy_env(object@env)
+  lapply(ls(object2@env), function(x) assign(x, get(x, object2@env), object@env))
+
+  object
```
pawelru, line 266: how about the following?
https://stackoverflow.com/questions/26057400/r-how-do-you-merge-combine-two-environments

we would omit copying objects from one env to another which could be time consuming and memory inefficient

gogonzo: very good suggestion!

gogonzo: I got one thing in mind. Imagine multiple reactives. 
- We start on top with `q` and we have two independent reactive outputs `q1`, `q2` which will have `q` as parent. 
- Then `q3` depends on `q2` and links another environment
- on the bottom if we join `q1` and `q3` we will need to figure out again all parents 

image

gogonzo: What I mean above is that in `q3` we don't want to link last environment to `q1` but we need to find first environment after the last shared environment. 
On the other hand "copying" objects is made currently by reference, which means that we only make a new binding but we don't change their pointers- they will change if we modify the object (for example filter or add column)

pawelru: I don't quite get below statements:

> two independent reactive outputs q1, q2 which will have q as parent

> Then q3 depends on q2 and links another environment

> we need to find first environment after the last shared environment

The problem here is as follows:
Does quosure preserve its parent env after evaluation?

This stems from the different understanding of eval method. My naive understanding could be summarised into the following:

```
eval <- function(obj, code) {
  ... # append code 
  obj@env <- with(obj@env, eval(code, obj@env))
}
```

This does not preserve any parent-child relationship of environments.

As a result - we are not able to find last shared env. 

Your example would be as follows:
```
(pseudocode)
q <- new_quosure()

q1 <- eval_code(q, x <- 1)
# q1@obj: env(x = 1)

q2 <- eval_code(q, y <- 2)
# q2@obj: env(y = 2)

q3 <- eval_code(q2, x <- 3)
# q3@obj: env(y = 2, x = 3)

qq <- join(q1, q3)
# qq@obj: env(y = 2, x = 3)
```

gogonzo: It means that if you change the parent of `q3` by:
```r
parent(q3@env) <- q1@env
``` 

then link between `q3` and `q2` will be lost, because `q2` will no longer be a parent of `q3`. It means that objects created in `q2` will no longer be in the joined environment because a env-tree would look like following:
```r
qq <- q1 <- q3
```

While it was
```
qq -- q1
   \_ q2 -- q3
```

pawelru: OK I got your point now and you are right - reference based joins won't work out. But I still believe that the quosure should not have a non-emptyenv parent env. I modified my idea into the following:

```
# new quo
q0 <- list2env(list(x = 0), parent = emptyenv())
rlang::env_print(q0)

# eval x = 1 in empty quo
#q1 <- list2env(list(x = 1), parent = q0)
q1 <- list2env(list(x = 1), parent = emptyenv()); rlang::env_coalesce(q1, q0)
for (i in c(q1, rlang::env_parents(q1))) rlang::env_print(i)
stopifnot(get("x", q1) == 1)
stopifnot(inherits(try(get("y", q1), silent = TRUE), "try-error"))
stopifnot(lobstr::obj_addr(get("x", q1)) != lobstr::obj_addr(get("x", q0)))

# eval y = 2 in empty quo
#q2 <- list2env(list(y = 2), parent = q0)
q2 <- list2env(list(y = 2), parent = emptyenv()); rlang::env_coalesce(q2, q0)
for (i in c(q2, rlang::env_parents(q2))) rlang::env_print(i)
stopifnot(get("x", q2) == 0)
stopifnot(get("y", q2) == 2)
stopifnot(lobstr::obj_addr(get("x", q2)) == lobstr::obj_addr(get("x", q0)))

# eval x = 3 in q2
#q3 <- list2env(list(x = 3), parent = q2)
q3 <- list2env(list(x = 3), parent = emptyenv()); rlang::env_coalesce(q3, q2)
for (i in c(q3, rlang::env_parents(q3))) rlang::env_print(i)
stopifnot(get("x", q3) == 3)
stopifnot(get("y", q3) == 2)
stopifnot(lobstr::obj_addr(get("x", q3)) != lobstr::obj_addr(get("x", q0)))
stopifnot(lobstr::obj_addr(get("y", q2)) == lobstr::obj_addr(get("y", q2)))

# join q3 into a q1
#qq <- list2env(as.list(q3), parent = q1)
qq <- new.env(parent = emptyenv()); rlang::env_coalesce(qq, q3); rlang::env_coalesce(qq, q1)
for (i in c(qq, rlang::env_parents(qq))) rlang::env_print(i)
stopifnot(get("x", qq) == 3)
stopifnot(get("y", qq) == 2)
stopifnot(lobstr::obj_addr(get("x", qq)) == lobstr::obj_addr(get("x", q3)))
stopifnot(lobstr::obj_addr(get("y", qq)) == lobstr::obj_addr(get("y", q2)))
```

commented-out code is my old and incorrect understanding
non-commented-out code is my new idea that seems to work on this dummy example

nikolas-burkoff:
After talking to Dawid:

The part I'm sort of a little concerned about is the implicitness of the passing of contents of environments between quosure objects: if this was chaining functions `g(y = f(x))` you would not expect `g(y)` to have access to local variables inside `f(x)` which we sort of do here as the quosure has access to what was in the environment(s) of all ancestor quosures.

I guess we are using nested structure instead as in `g <- function() {  temp <- 5;  h <- function(x) { temp * x } ...}`

Why is this an issue? - we're interested in the output of `h` not just `g` also it makes it much harder to reason about what's available inside the environments especially as the structure of the nesting and their contents itself is reactive (i.e. dependent on input$x you may have a different nested structure) and the nesting is implicit and scattered throughout the shiny module. Also as everything is sort of hidden inside environments it makes it much harder to debug (and formally test) and learn

Maybe there isn't a way round this though

R/Quosure.R
gogonzo: Ok, I see your point. By non-empty you probably mean `parent.env(.GlobalEnv)` because if we stem from `emptyenv` we won't have access to any of the packages namespaces. I will use `rlang::env_coalesce` instead of `for` loop to "merge" environments.

gogonzo:
@nikolas-burkoff @pawelru 
what about concept of `reactive_quosure` being a wrapper on `quosure`. It would be something like:

```r
tmg_srv <- function(data, ...) {
  q1 <- new_reactive_quosure(data) # has ADSL and ADTTE
  
  q2 <- reactive_quosure(depends = q1, {
    ANL <- left_join(ADSL, ADTTE)
  })
  
  q3 <- reactive_quosure(depends = q2, {
    counts <- ANL %>% group_by(!!input$group_var) %>% summarize(n = n())
  })
  
  observeEvent(q3()[["counts"]], {
    print(q3()[["counts"]])
  })
```

R/Quosure.R
```diff
@@ -0,0 +1,290 @@
+#' Reproducible class with environment and code.
+#'
+#' @slot code (`character`) to reproduce the environment
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#' of the `code` slot.
+setClass(
+  "Quosure",
+  representation(code = "character", env = "environment"),
+  prototype(code = character(0), env = new.env(parent = parent.env(.GlobalEnv)))
```
Polkas, line 9: Please be aware that any library will not be attached to such environment we have to use sth like here https://github.com/insightsengineering/teal.data/blob/76c2e33f59057215569be210b54a10d28bc4f476/R/TealDatasetConnector_constructors.R#L827-L837

Example:

```r
ee <- new.env(parent = parent.env(globalenv()))
eval(quote({library(pacs); pac_description}), ee)

# VS

ee <- new.env(parent = parent.env(globalenv()))
ee$library <- function(...) {
  mc <- match.call()
  mc[[1]] <- quote(base::library)
  eval(mc, envir = globalenv())
  this_env <- parent.frame()
  if (!identical(this_env, globalenv())) {
    parent.env(this_env) <- parent.env(globalenv())
  }
}
eval(quote({library(cat2cat); cat2cat}), ee)
``

File: R/Quosure.R (hunk 10 above)
Comment (Polkas) [reply R/Quosure.R:271]: We have to use parent.env(.GlobalEnv) and it have to be dynamic to take into account any library called internally.
To be more precious  parent.env(.GlobalEnv) provided to us all base packages. Later when we execute any code with library call we have to update such environment to once more be one level below the Globalenv so we have access to all needed attached packages.

**Comment by nikolas-burkoff** [comment]
Should we have an `ls` function for quosures 

`as.list(q@env))` is ok I guess

File: R/Quosure.R (hunk 11 above)
Comment (nikolas-burkoff) [reply R/Quosure.R:9]: I wonder if library calls should be treated differently (or even not allowed)...

File: R/Quosure.R [hunk 12]
```diff
@@ -0,0 +1,310 @@
...
+
+
+#' @export
+as_quosure <- function(x) {
+  UseMethod("as_quosure")
+}
+
+#' @export
+as_quosure.list <- function(x) {
+  new_quosure(list2env(x), code = as.expression(attr(x, "code")))
```
Polkas, line 309: as.expression for code is not needed as it will be automatically properly dispatch

Polkas:
1) Can we use a simpler representation
2) Apply quosure to a tmc module
3) name instead of unique id -  should we care about informative names
4) Do we need comments in quosures  etc. or can we only support a pushing language - but not support other types
5) Explain to Nik why this works 

q1 <- new_quosure(c("x <- 5", "y <- 10", "x <- 7"), env = new.env())
q2 <- new_quosure(c("x <- 5", "y <- x"), env = new.env())

join(q1, q2)

nikolas-burkoff:
For me:

1) I think we do need an object structure, manipulating lists is too easy to go wrong, and as we see there's all sorts of complexities we want hidden from the users of quosures 

2) looks straightforward (though I would still argue about https://github.com/insightsengineering/teal.code/pull/49#issuecomment-1196755057) 

3) I don't care about informative names - I don't mind about debug options but what I would like is the SRC to show the failing code at the moment (at least in tmg) it doesn't so it makes it harder to debug

4) Supporting other types - don't see why not, as long as the underlying structure stores them all in one type - as for comments, this depends on the user requirements - if this is meant to be a script people are taking to use themselves and adapt then I think we should support them, if this is a script which is either never ran but their as traceability or only ever ran blindly (i.e. by a pipeline) then no need
One thing we should think about here and in point 3) is how the SRC relates to the reporting - you could imagine wanting the code from different parts to be placed in different places in in the report in the future...

5) Yeah this is wrong and should be fixed...

Polkas:
1. Quosure class is ok for me. This Structure is mostly a middleware and we assume It is not used directly.
2. 
3. The former id could work as a guard of writing a proper code pipeline. When sb think about a proper id and then could print all of them , this could work as a validation or confirmation step. I do not care about Debugger if the Show R Code fail messages will be informative.
4. we are passing a language to the eval method which could be evaluated directly but we convert it to character to then parse and evaluate. THis for sure not looks a proper move. now the idea to support only language (or only string) have another reason. But if the cycle will end on the  luanagage and string will be parsed just after it was provided I think we could support all types.
5. Yes, combining the qusures should be rethinked . In my opinion we should remove https://github.com/insightsengineering/teal.code/blob/a3567fb76d198ae7d8c44ab8e0401d3821cb8c6c/R/Quosure.R#L246-L267  or rewrite it to sth more sophisticated.

R/Quosure.R
gogonzo: Yes, not needed, thanks. I put this once just to test something

gogonzo:
> DUALITY OF DATA RESOURCE. I have a feeling it is a little be complex as we have an additional place to handle when interacting https://github.com/insightsengineering/teal.modules.clinical/pull/542/files Normally we play with anl_merged and now we use it still in half places and in other half we access the Quosure (for raw data). With @nikolas-burkoff we feel we don't have to extract the data from Quosures in most cases and we could use directly data argument. The only change is that from Quosures we are getting preprocessing but in the old code we were not using it so should not be a problem.

Yup, I think also that places where we had `datasets$get_data(data, filtered = TRUE)` can be replaced by `data[[<dataname>]]()` instead of `q[[<dataname>]]`. For ANL we have no choice because `q` is the only object where we can extract data from.
Regarding Nik's [comment](https://github.com/insightsengineering/teal.code/pull/49#issuecomment-1196755057)  - if we could only use the latest it would mean that we can't make `plot` and `table` and use both outputs somewhere. More complex would be adding calls. For example

```r
q1 <- eval_code(q, data <- <some data>)
q2 <- eval_code(q1, data_2 <- <summary of data>)
q3 <- eval_code(q2, names(data)[1] <- "renamed first column")
q4 <- eval_code(q3, data_merged <- merge(data, data2)) 
```

If each quosure evaluation would return one object this would happen:
- q2 have only data_2
- in eval_code for q3 object `data` is not available because q2 has only data_2. Even if we do this somehow this quosure would have only `names(data)[1]` which will break the next line

I don't see the point of having quosure which keeps only one (last) object. If we go this way, probably we will need to make some `reactuveQuosure` method which would be like a `reactive`. But then we will need to find the way to separate a `quosure` code from the shiny code.

For example:

```r
anl_merged <- teal.transform::merge_expression_module(...)

r1 <- reactive(<some data>)

q1 <- reactive(new_quosure(...))

q2 <- reactive_quosure(depends_on = q1, !!anl_merged()$expr)

q3 <- reactive_quosure(depends_on = q2, {
  some_data <- r1()
  some_statistics <- if (input$some_switch) {
     <some computation>
  } else {
    
    <some other computation>
  }
})


r4 <- reactive({ q3() })

observeEvent(q3(), {})
``` 

Please look at the code above. We can make a better version of `metaReactive`. Better means that each `q` is a normal reactive which returns quosure, this means that it can trigger observeEvent, it can be used in other reactive (r4), etc.
Problem is in the `q3`. In this case we don't want to show the whole if but only one of the option. Becase `reactive_quosure` is a reactive this means that we can "evaluate" other reactives there (see `some_data <- r1()` in `q3`). We don't want to include any shiny code in our "reproducible" code. I assume this code separation made shinymeta so difficult to implement and so limited now to use. If we could solve this problem It's a ticket to paradise.

> NOT CARE ABOUT DEBUGGER. The former id could work as a guard of writing a proper code pipeline. When sb think about a proper id and then could print all of them , this could work as a validation or confirmation step. I do not care about Debugger if the Show R Code fail messages will be informative. Currently in tmg there are not such fail messages.

Let's discuss this in details if we want to have some sort of "eval details" or do we need to keep R code and some styled error message. In all cases we need to `tryCatch` and keep ConditionMessages in the class if any.

> JOIN PROBLEMS. Yes, combining the qusures should be rethinked . In my opinion we should remove

Let's solve this 

> Additional curly brackets {} in the show R code
It's a problem of the modules not quosure - we pass the code with the bracket so it's reflected in the output. We can re-style the code in tm modules a little bit.

> NEST-like design which e.g. assume the code attr is not evaluated as is already applied.
What do you mean?

> Supporting other types - don't see why not, as long as the underlying structure stores them all in one type - as for comments, this depends on the user requirements - if this is meant to be a script people are taking to use themselves and adapt then I think we should support them, if this is a script which is either never ran but their as traceability or only ever ran blindly (i.e. by a pipeline) then no need
One thing we should think about here and in point 3) is how the SRC relates to the reporting - you could imagine wanting the code from different parts to be placed in different places in in the report in the future...

It means that we need a way to extract specific code elements and named elements are convenient.

gogonzo:
Hi @kumamiao , please let us know here about importance of the possibility to add comments and empty lines in the "show R code".
We consider simplifying the "next chunks" class by disabling code being a `character` and keeping it as `language`/`expression`. It's a minor simplification, but if comments/line breaks are not needed it might be worth to simplify.

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
+#' Reproducible class with environment and code.
+#'
+#' @slot code (`character`) to reproduce the environment
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#' of the `code` slot.
+setClass(
```
Polkas, line 6: ?setClass

validity: supplied a validity-checking method for objects from this class. For clearer code, use a separate call to setValidity().
...
prototype: supplies an object with the default data for the slots in this class. A more flexible approach is to write a method for initialize().
...
representation is an argument inherited from S that included both slots and contains, but the use of the latter two arguments is clearer and recommended.

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  # insert (and overwrite) objects from env2 to env
+  x@env <- .copy_env(x@env)
+  lapply(ls(y@env), function(xx) assign(xx, get(xx, y@env), x@env))
+
+  x
+})
+
+#' If two `QCode` can be joined
+#'
+#' Checks if two `QCode` objects can be combined.
```
Polkas, line 275:
```suggestion
#' If two `Quosure` can be joined
#'
#' Checks if two `Quosure` objects can be combined.
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  standardGeneric("check_joinable")
+})
+
+setMethod(
+  "check_joinable", signature(x = "Quosure", y = "Quosure"),
+  function(x, y) {
+    common_names <- intersect(ls(x@env), ls(y@env))
+    is_overwritten <- vapply(common_names, function(xx) {
+      !identical(get(xx, x@env), get(xx, y@env))
+    }, logical(1))
```
Polkas, line 292:
```suggestion
    is_overwritten <- vapply(common_names, function(el) {
      !identical(get(el, x@env), get(el, y@env))
    }, logical(1))
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+    shared_in_x <- match(shared_ids, x@id)
+    shared_in_y <- match(shared_ids, y@id)
+
+    # indices of shared ids should be 1:n in both slots
+    if (identical(shared_in_x, shared_in_y) && identical(shared_in_x, seq_along(shared_ids))) {
+      TRUE
+    } else if (!identical(shared_in_x, shared_in_y)) {
+      paste(
+        "Common code of joined objects doesn't have the same indices. Thismeans that `x` and `y` have ",
+        "can't be joined together as it's not possible to determine order of the evaluation.",
```
Polkas, line 315:
```suggestion
        "The common code of joined objects doesn't have the same indices. It means that `x` and `y`",
        "can't be joined together as it's impossible to determine the evaluation's order.",
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+      )
+    }
+
+    shared_ids <- intersect(x@id, y@id)
+    if (length(shared_ids) == 0) return(TRUE)
+
+    shared_in_x <- match(shared_ids, x@id)
+    shared_in_y <- match(shared_ids, y@id)
+
+    # indices of shared ids should be 1:n in both slots
```
Polkas, line 309:
```suggestion
    # indices of shared ids should be 1:n in both slots
    # beginning of the code, like loading libraries and data
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+        "Common code of joined objects doesn't have the same indices. Thismeans that `x` and `y` have ",
+        "can't be joined together as it's not possible to determine order of the evaluation.",
+        collapse = ""
+      )
+    } else {
+      paste(
+        "Common code of joined object doesn't start from index = 1. This means that joined object(x) have ",
+        "some extra code elements before.",
+        collapse = ""
+      )
```
Polkas, line 323:
```suggestion
      paste(
        "The common code of joined object does not start from index = 1.",
        "It means that joined object(x) has some extra code elements before.",
        collapse = ""
      )
```

kumamiao:
> Hi @kumamiao , please let us know here about importance of the possibility to add comments and empty lines in the "show R code".

I looked at the guidance on submitting TLG code, and it does look like well-commented/indented and structured code is the recommended practice to readable code. So, if feasible, it would be great to still include the comments and empty lines. Thanks!

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
+#' Reproducible class with environment and code.
+#'
+#' @slot code (`character`) to reproduce the environment
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#' of the `code` slot.
+setClass(
+  "Quosure",
+  representation(env = "environment", code = "character", id = "integer"),
```
Polkas, line 8:
```suggestion
  slots = c(env = "environment", code = "character", id = "integer"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+      return("@code and @id slots must have the same length.")
+    }
+    if (any(duplicated(object@id))) {
+      return("@id contains duplicated values.")
+    }
+
+    TRUE
+  }
+)
+
```
Polkas, line 21:
```suggestion
  prototype(env = new.env(parent = parent.env(.GlobalEnv)), code = character(0), id = integer(0))
)

setValidity("Quosure",  function(object) {
    if (length(object@code) != length(object@id)) {
      "@code and @id slots must have the same length."
    } else if (any(duplicated(object@id))) {
     "@id contains duplicated values."
    } else {
      TRUE
    }
})
  
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+    if (length(object@code) != length(object@id)) {
+      return("@code and @id slots must have the same length.")
+    }
+    if (any(duplicated(object@id))) {
+      return("@id contains duplicated values.")
+    }
+
+    TRUE
+  }
+)
```
Polkas, line 20:
```suggestion
setClass(
  "Quosure",
  slots = c(env = "environment", code = "character", id = "integer"),
  prototype = list(env = new.env(parent = parent.env(.GlobalEnv)), code = character(0), id = integer(0))
)

#' It takes a `Quosure` class and returns TRUE if the input is valid
#' @keywords internal
setValidity("Quosure", function(object) {
  if (length(object@code) != length(object@id)) {
    "@code and @id slots must have the same length."
  } else if (any(duplicated(object@id))) {
    "@id contains duplicated values."
  } else {
    TRUE
  }
})
```

gogonzo:
@kumamiao Thanks! So, we are going to keep code as character in the class.

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  id_unique <- !y@id %in% x@id
+  x@id <- c(x@id, y@id[id_unique])
+  x@code <- .keep_code_name_unique(x@code, y@code[id_unique])
+
+  # insert (and overwrite) objects from env2 to env
+  x@env <- .copy_env(x@env)
+  lapply(ls(y@env), function(xx) assign(xx, get(xx, y@env), x@env))
+
+  x
+})
```
Polkas, line 271:
```suggestion
#' Join two `Quosure` objects
#'
#' Combine two `Quosure` object by merging their environments and the code.
#' Not every `Quosure` objects can be combined:
#' - if their environments contains objects of the same name but not identical
#' - if `y` has unique code element placed before common element. This means that `y`
#' in the environment of the `y` was evaluated some extra code before which can influence
#' reproducibility
#' - more cases todo
#' @param x (`Quosure`)
#' @param y (`Quosure`)
#' @param resolve (`logical`) if we want to remove duplicate code if possible, by default `TRUE`.
#' @export
setGeneric("join", function(x, y, resolve) {
  standardGeneric("join")
})

setMethod("join", signature("Quosure", "Quosure", "ANY"), function(x, y, resolve) {
  if (missing(resolve) || !is.logical(resolve)) resolve <- TRUE
  if (resolve) {
    if (!isTRUE(check_joinable(x, y))) {
      stop(join_validation)
    }
    id_unique <- !y@id %in% x@id
    x@id <- c(x@id, y@id[id_unique])
    x@code <- .keep_code_name_unique(x@code, y@code[id_unique])
  } else {
    x@id <- c(x@id, y@id)
    x@code <- .keep_code_name_unique(x@code, y@code)
  }
  # insert (and overwrite) objects from env2 to env
  x@env <- .copy_env(x@env)
  lapply(ls(y@env), function(xx) assign(xx, get(xx, y@env), x@env))
  x
})
```

Polkas:
New propositions tested with:

``` r
q0 <- new_quosure(c("x <- 5"),
                  env = list2env(list(x = 5)))
#> Error in new_quosure(c("x <- 5"), env = list2env(list(x = 5))): could not find function "new_quosure"
q1 <- new_quosure(c("x <- 5", "y <- 10", "x <- 7"),
                  env = list2env(list(x = 7, y = 10)))
#> Error in new_quosure(c("x <- 5", "y <- 10", "x <- 7"), env = list2env(list(x = 7, : could not find function "new_quosure"
q2 <- new_quosure(c("x <- 5", "z <- x", "x <- 7"),
                  env = list2env(list(x = 7, z = 5)))
#> Error in new_quosure(c("x <- 5", "z <- x", "x <- 7"), env = list2env(list(x = 7, : could not find function "new_quosure"

q3 <-  q0 |> eval_code(z <- 5)
#> Error in eval_code(q0, z <- 5): could not find function "eval_code"

q4 <-  q0 |> eval_code(t <- 5)
#> Error in eval_code(q0, t <- 5): could not find function "eval_code"

q5 <-  q0 |> eval_code(x <- 7)
#> Error in eval_code(q0, x <- 7): could not find function "eval_code"

join(q0, q1) # the same error for join(q0, q2)
#> Error in join(q0, q1): could not find function "join"

get_code(join(q1, q2)) # all
#> Error in get_code(join(q1, q2)): could not find function "get_code"

get_code(join(q0, q3))
#> Error in get_code(join(q0, q3)): could not find function "get_code"

get_code(join(q3, q4))
#> Error in get_code(join(q3, q4)): could not find function "get_code"

get_code(join(q3, q5)) # fail
#> Error in get_code(join(q3, q5)): could not find function "get_code"

get_code(join(q3, q5, resolve = FALSE)) # succcess
#> Error in get_code(join(q3, q5, resolve = FALSE)): could not find function "get_code"
```

<sup>Created on 2022-08-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+#'
+#' @export
+setGeneric("new_quosure", function(env = new.env(parent = parent.env(.GlobalEnv)), code = character(0)) {
+  standardGeneric("new_quosure")
+})
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(env = "environment", code = "character"),
```
Polkas, line 40:
```suggestion
  "new_quosure", signature = c(env = "environment", code = "character"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+    id <- sample.int(.Machine$integer.max, size = length(code))
+
+    methods::new("Quosure", env = env, code = code, id = id)
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(env = "environment", code = "expression"),
```
Polkas, line 58:
```suggestion
  "new_quosure", signature = c(env = "environment", code = "expression"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  function(env, code) {
+    code_char <- as.character(code)
+    new_quosure(env = env, code = code_char)
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(env = "environment", code = "language"),
```
Polkas, line 68:
```suggestion
  "new_quosure", signature = c(env = "environment", code = "language"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  function(env, code) {
+    code_expr <- as.expression(code)
+    new_quosure(env = env, code = code_expr)
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(env = "environment", code = "ANY"),
```
Polkas, line 78:
```suggestion
  "new_quosure", signature = c(env = "environment", code = "ANY"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  function(env, code) {
+    quoted_expr <- substitute(code)
+    new_quosure(env = env, code = quoted_expr)
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(env = "list"),
```
Polkas, line 88:
```suggestion
  "new_quosure", signature = c(env = "list"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+      })
+    }
+    new_quosure(env = list2env(env), code = code)
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure", signature(code = "missing", env = "missing"),
```
Polkas, line 118:
```suggestion
  "new_quosure", signature = c(code = "missing", env = "missing"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+#'
+#' @export
+setGeneric("eval_code", function(object, code, name = "code") {
+  standardGeneric("eval_code")
+})
+
+#' @rdname eval_code
+#' @export
+setMethod(
+  "eval_code", signature("Quosure", "character"),
```
Polkas, line 144:
```suggestion
  "eval_code", signature = c("Quosure", "character"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+    eval(parse(text = code), envir = object@env)
+    lockEnvironment(object@env)
+    object
+  }
+)
+
+#' @rdname eval_code
+#' @export
+setMethod(
+  "eval_code", signature("Quosure", "expression"),
```
Polkas, line 167:
```suggestion
  "eval_code", signature = c("Quosure", "expression"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  function(object, code, name) {
+    code_char <- as.character(code)
+    eval_code(object, code_char, name = name)
+  }
+)
+
+#' @rdname eval_code
+#' @export
+setMethod(
+  "eval_code", signature("Quosure", "language"),
```
Polkas, line 177:
```suggestion
  "eval_code", signature = c("Quosure", "language"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+  function(object, code, name) {
+    code_char <- as.expression(code)
+    eval_code(object, code_char, name = name)
+  }
+)
+
+#' @rdname eval_code
+#' @export
+setMethod(
+  "eval_code", signature("Quosure", "ANY"),
```
Polkas, line 187:
```suggestion
  "eval_code", signature = c("Quosure", "ANY"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+#' @name get_var
+#' @export
+setGeneric("get_var", function(object, var) {
+  standardGeneric("get_var")
+})
+
+
+#' @rdname get_var
+#' @export
+setMethod("get_var", signature("Quosure", "character"), function(object, var) {
```
Polkas, line 208:
```suggestion
setMethod("get_var", signature = c("Quosure", "character"), function(object, var) {
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+#' - indices of the shared code are not consecutive or don't start from 1
+#' @param x (`Quosure`)
+#' @param y (`Quosure`)
+#' @return logical(1)
+setGeneric("check_joinable", function(x, y) {
+  standardGeneric("check_joinable")
+})
+
+setMethod(
+  "check_joinable", signature(x = "Quosure", y = "Quosure"),
```
Polkas, line 287:
```suggestion
  "check_joinable", signature = c(x = "Quosure", y = "Quosure"),
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+#' @return named `character` with the reproducible code.
+#' @export
+setGeneric("get_code", function(object) {
+  standardGeneric("get_code")
+})
+
+#' Get the code from the `Quosure`
+#'
+#' @param object (`Quosure`)
+setMethod("get_code", signature("Quosure"), function(object) {
```
Polkas, line 234:
```suggestion
setMethod("get_code", signature = "Quosure", function(object) {
```

R/Quosure.R
```diff
@@ -0,0 +1,345 @@
...
+#' reproducibility
+#' - more cases todo
+#' @param x (`Quosure`)
+#' @param y (`Quosure`)
+#' @export
+setGeneric("join", function(x, y) {
+  standardGeneric("join")
+})
+
+setMethod("join", signature("Quosure", "Quosure"), function(x, y) {
```
Polkas, line 254:
```suggestion
setMethod("join", signature = c("Quosure", "Quosure"), function(x, y) {
```

Polkas:
linked to https://github.com/insightsengineering/teal.code/issues/57

tests/testthat/test-quosure-constructor.R
```diff
@@ -0,0 +1,76 @@
...
+testthat::test_that("new_quosure works with code being expression", {
+  env <- new.env()
+  env$iris1 <- iris
+  q <- new_quosure(parse(text = "iris1 <- iris"), env = env)
+  testthat::expect_identical(q@env, env)
+  testthat::expect_identical(q@code, c(`initial code` = "iris1 <- iris"))
+  testthat::expect_true(checkmate::test_int(q@id))
+})
+
+testthat::test_that("new_quosure works with code a quoted expression", {
```
mhallal1, line 31:
```suggestion
testthat::test_that("new_quosure works with code being quoted expression", {
```

tests/testthat/test-quosure-join.R
```diff
@@ -0,0 +1,152 @@
...
+
+  testthat::expect_equal(
+    as.list(q@env),
+    list(iris1 = iris, iris2 = iris, mtcars1 = mtcars, mtcars2 = mtcars)
+  )
+
+  testthat::expect_identical(q@id, c(q1@id, q2@id[3]))
+})
+
+testthat::test_that("Not possible to join quosures which shares some code and modify one of the shared object", {
```
mhallal1, line 91:
```suggestion
testthat::test_that("Not possible to join quosures which share some code when one of the shared object was modified", {
```

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+  }
+  env_new
+}
+
+.keep_code_name_unique <- function(x, y = character(0)) {
+  combined <- c(x, y)
+  if (length(names(combined)) == 0) {
+    return(combined)
+  }
+  setNames(combined, make.unique(names(combined)))
```
nikolas-burkoff, line 365: stats::setNames

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+  } else {
+    TRUE
+  }
+})
+
+#' Initialize `Quosure` object
+#'
+#' Initialize `Quosure` object with `code` and `env`. In order to have `Quosure` reproducible
+#' one needs to initialize with `env` which can be reproduced by the `code`. Alternatively, one
+#' can create an empty `Quosure` and evaluate the expressions in this object.
```
nikolas-burkoff, line 34: Using `eval_code`

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+#' @name new_quosure
+#'
+#' @param code (`character(1)` or `language`) code to evaluate. Accepts and stores comments also.
+#' @param env (`environment`) Environment being a result of the `code` evaluation.
```
nikolas-burkoff, line 38: So it's not quite this - as there's nothing actually stopping anyone doing `q <- new_quosure(env = list2env(list(x = 5)), code = x <- 10)` it's more correctly the code which will be prepended to the `quosure::get_code` code and in order for reproducibility it should create the contents of `env`

(I wonder if we should have some tests showing this is ok - and also should we have an issue to have a check argument in the future so that teal.data could use quosures....)

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+})
+
+#' Initialize `Quosure` object
+#'
+#' Initialize `Quosure` object with `code` and `env`. In order to have `Quosure` reproducible
+#' one needs to initialize with `env` which can be reproduced by the `code`. Alternatively, one
+#' can create an empty `Quosure` and evaluate the expressions in this object.
+#' @name new_quosure
+#'
+#' @param code (`character(1)` or `language`) code to evaluate. Accepts and stores comments also.
```
nikolas-burkoff, line 37: Can we have tests for this...

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+#' @param env (`environment`) Environment being a result of the `code` evaluation.
+#'
+#' @export
```
nikolas-burkoff, line 40: We should have some examples

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+    stop(join_validation)
+  }
+
+  id_unique <- !y@id %in% x@id
+  x@id <- c(x@id, y@id[id_unique])
+  x@code <- .keep_code_name_unique(x@code, y@code[id_unique])
+
+  # insert (and overwrite) objects from env2 to env
+  x@env <- .copy_env(x@env)
+  lapply(ls(y@env), function(xx) assign(xx, get(xx, y@env), x@env))
```
nikolas-burkoff, line 288: Note that `ls()` has this fun quirk:

image

(If you use all.names = TRUE then you will get .random.seed if it exists - we may also want to set sorted = FALSE)

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+    )
+  }
+}
+
+#' Copy environment content to the other environment
+#' To avoid sharing the same environment by multiple `Quosure` objects
+#' @keywords internal
+.copy_env <- function(env) {
+  # need to copy the objects from old env to new env
+  # to avoid "chunks" sharing the same environment
```
nikolas-burkoff, line 352: chunks...

R/Quosure.R
```diff
@@ -0,0 +1,366 @@
...
+  standardGeneric("eval_code")
+})
+
+#' @rdname eval_code
+#' @export
+setMethod(
+  "eval_code",
+  signature = c("Quosure", "character"),
+  function(object, code, name) {
+    checkmate::assert_character(name, len = 1L)
```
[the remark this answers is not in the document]
nikolas-burkoff: If NA is ok, we should probably have explicit tests for it

tests/testthat/test-quosure-join.R
```diff
@@ -0,0 +1,152 @@
...
+  testthat::expect_s4_class(cq, "Quosure")
+  testthat::expect_equal(cq@env, list2env(list(a1 = 1, b1 = 2, a2 = 3)))
+  testthat::expect_identical(
+    cq@code,
+    c(`initial code` = "a1 <- 1", `code` = "a2 <- 3", `code.1` = "b1 <- 2")
+  )
+  testthat::expect_identical(cq@id, union(q1@id, q2@id))
+})
+
+testthat::test_that("Quosure objects aren't mergeable if they share common code proceeded with some other code", {
```
nikolas-burkoff, line 132: So this is not quite true, it's not common code, it's common quosure as here you can join `q5` and `q6` even though they both have common _code_ `c1 <- 1` proceeded by some other code

```
q1 <- new_quosure(code = "a1 <- 1", env = list2env(list(a1 = 1)))
q2 <- new_quosure(code = "b1 <- 1", env = list2env(list(b1 = 1)))

q3 <- new_quosure(code = "c1 <- 1", env = list2env(list(c1 = 1)))
q4 <- new_quosure(code = "c1 <- 1", env = list2env(list(c1 = 1)))


q5 <- join(q1, q3)
q6 <- join(q2, q4)
```

I think we need to be really careful when we mean "code" and when we mean "quosure" throughout the PR

nikolas-burkoff (commented):
A few comments inline
- Roxygen examples please
- We should add documentation vignette as an issue?
- Should we soft deprecate chunks in this PR as it's going to a feature branch
- I think we need better test coverage - testing 
    - eval_code/get_code 
    - get_var ([[]]) 
    - constructor with reactives inside list

tests/testthat/test-eval_code.R
```diff
@@ -0,0 +1,97 @@
...
+  testthat::expect_identical(parent.env(q1@env), parent.env(.GlobalEnv))
+
+  q2 <- eval_code(q1, "a <- 1L")
+  testthat::expect_identical(parent.env(q2@env), parent.env(.GlobalEnv))
+  q3 <- eval_code(q2, "b <- 2L")
+  testthat::expect_identical(parent.env(q3@env), parent.env(.GlobalEnv))
+})
+
+testthat::test_that("library have to be called separately before using function from package", {
+  q1 <- eval_code(new_quosure(), "library(checkmate)")
```
nikolas-burkoff, line 25: Is the state before and after running this test definitely the same?

gogonzo: content of the @env is the same but location of the @env changes.

Normally this is the environment hierarchy
```
- empty env
- base_env
- package1 env
- packageN env
- .GlobalEnv
```

When we create a quosure env is a parent of .GlobalEnv

```
- empty env
- base_env
- packageA env
- .... other packages env
- packageX env
- @env
- .GlobalEnv
```

When we call `library(Y)` within the eval_code, then Y becomes a parent of .GlobalEnv
```
- empty env
- base_env
- package A env
- .... other packages env
- package X env
- @env
- package Y env
- .GlobalEnv
```

R/Quosure.R
gogonzo: `parent.env(object@env) <- parent.env(.GlobalEnv)` is applied in `new_quosure` and in `eval_code`. It's important that `eval_code` keeps updates parent in each evaluation - then modification caused by `library()` calls doesn't influence quosure so much.

Polkas:
Approve from me to merge to the feature branch:)

PR #66, opened 2022-09-15
Teal refactor

pawelru:
Part of https://github.com/insightsengineering/teal/issues/731


R/quosure-utils.R
```diff
@@ -0,0 +1,28 @@
...
+
+#' Combine two code vectors and keep unique names
+#'
+#' Combines two code vectors and modifies duplicated names.
+#' @param x (`character`) vector of code.
+#' @param y (`character`) vector of code.
+#' @return `character` containing both `x` and `y` combined with
+#'  unique names.
+#' @keywords internal
+.keep_code_name_unique <- function(x, y = character(0)) {
```
pawelru, line 22: keep_name_unique - this name indicates that the main responsibility is to _mutate names_ whereas the main responsibility here is to make _combine_ action.
I suggest to either rename it into c_with_unique_names or split it into two steps: (i) c, (ii) keep unique names.

R/quosure-constructor.R
```diff
@@ -0,0 +1,103 @@
...
+
+    if (checkmate::test_list(env, "reactive")) {
+      env <- lapply(env, function(x) {
+        if (inherits(x, "reactive")) {
+          x()
+        } else {
+          x
+        }
+      })
+    }
```
pawelru, line 90: this scenario is neither documented nor covered by tests

R/quosure-constructor.R
```diff
@@ -0,0 +1,103 @@
...
+#' @export
+setMethod(
+  "new_quosure",
+  signature = c(env = "list"),
+  function(env, code) {
+    if (!missing(code)) {
+      warning(
+        "`code` argument is ignored when creating Quosure from the list.",
+        "\nPlease pass the code througn attr(<list>, 'code')"
+      )
```
pawelru, line 74: why's that? `list` and `environment` are very close structures and the approach for passing on the code information differs.

R/quosure-constructor.R
```diff
@@ -0,0 +1,103 @@
...
+    code_expr <- as.expression(code)
+    new_quosure(env = env, code = code_expr)
+  }
+)
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure",
+  signature = c(env = "list"),
```
pawelru, line 68: I am not so sure if we should support constructor from `list`. Please note that a lot of classes (including non-base ones) inherits from a `list`, e.g. `data.table`, `R6`, etc. Allowing `list` opens unknown bag of issues. We should start small and allow environments only.

R/quosure-eval_code.R
```diff
@@ -0,0 +1,68 @@
...
+      code <- paste(code, collapse = "\n")
+      names(code) <- name
+    }
+    id <- sample.int(.Machine$integer.max, size = length(code))
+    object@id <- c(object@id, id)
+    object@code <- .keep_code_name_unique(object@code, code)
+
+    # need to copy the objects from old env to new env
+    # to avoid updating environments in the separate objects
+    object@env <- .copy_env(object@env)
```
pawelru, line 40: I don't really get why it's needed. Can you give a simple example?

R/quosure-utils.R
```diff
@@ -0,0 +1,28 @@
+#' Copy environment content to the other environment
+#' To avoid sharing the same environment by multiple `Quosure` objects
+#' @keywords internal
+.copy_env <- function(env) {
```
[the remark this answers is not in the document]
gogonzo: `.copy_env` copies the environment object to the new environment which is the independent of the .GlobalEnv `new.env(parent = parent.env(.globalEnv))`. If we don't do this following code would not produce expected results

https://github.com/insightsengineering/teal.code/blob/a5c454a5ab1f89173af563c39694922aad4a5f61/tests/testthat/test-eval_code.R#L7

R/quosure-eval_code.R
gogonzo: Please see the comment here:

https://github.com/insightsengineering/teal.code/pull/66#discussion_r971872869

R/quosure-constructor.R
gogonzo: This was introduces to simplify `Quosure` constructor in the modules. We debate about `data` argument in the modu
```r
function(data <reactive list with attr>) {
  # instead of 
  new_quosure(sapply(data, function(x) x()), code = attr(data, "code"))

  # we can use this
  new_quosure(data)
}
```

Since we are going to change "data" to some class (let's call it "elephant") we will remove this constructor and introduce `elephant.get_data` and `elephant.get_code` and we will call `new_quosure(get_data(data), get_code(data))`

R/quosure-join.R
```diff
@@ -0,0 +1,103 @@
...
+    stop(join_validation)
+  }
+
+  id_unique <- !y@id %in% x@id
+  x@id <- c(x@id, y@id[id_unique])
+  x@code <- .keep_code_name_unique(x@code, y@code[id_unique])
+
+  # insert (and overwrite) objects from env2 to env
+  x@env <- .copy_env(x@env)
+  lapply(ls(y@env, all.names = TRUE), function(xx) assign(xx, get(xx, y@env), x@env))
```
pawelru, line 46: [`rlang::env_coalesce`](https://rlang.r-lib.org/reference/env_clone.html)

R/quosure-join.R
```diff
@@ -0,0 +1,103 @@
...
+  if (!isTRUE(join_validation)) {
+    stop(join_validation)
+  }
+
+  id_unique <- !y@id %in% x@id
+  x@id <- c(x@id, y@id[id_unique])
+  x@code <- .keep_code_name_unique(x@code, y@code[id_unique])
+
+  # insert (and overwrite) objects from env2 to env
+  x@env <- .copy_env(x@env)
```
pawelru, line 45: why do we need to copy env from self into self?

gogonzo: imagine you have the situation like this:
```r
x <- new_quosure(env = list2env(a = 1), code = "a <- 1")
y <- new_quosure(env = list2env(a = 2), code = "a <- 2") 

z <- join(x, y)
```

If we didn't copy the environment we join would result with `z` and `x` sharing the same environment (same as in eval_code).

R/quosure-constructor.R
pawelru: OK I understand the need but IMO this should be handled on the caller side not to pollute this general functionality. Your example should be then: `new_quosure(list2env(get_data(data)), get_code(data))`

Coupling functionalities is also a concern. Please note that the idea of code as an attribute comes from somewhere else and it affects this package.

R/quosure-join.R
gogonzo: I think you are right here and comment from 44 is not valid as `check_joinable` prevent to join when one quosure modified any of the common objects.

R/quosure-constructor.R
gogonzo: This will be handled on a caller side soon

R/quosure-show.R
```diff
@@ -0,0 +1,18 @@
...
+#' q1
+#' @export
+setMethod("show", "Quosure", function(object) {
+  obs <- names(as.list(object@env))
+  if (length(obs) > 0) {
+    cat(paste("A quosure object containing:", paste(obs, collapse = ", ")))
+  } else {
+    cat("A quosure object containing no objects")
+  }
+})
```
pawelru, line 18: 1. for `cat()` calls we need to put `\n` at the very end
```
> 1
[1] 1
> q
A quosure object containing no objects> 2             # <- here!
[1] 2
```

2. How about utilizing `renv::print_env`? I recently discovered this and I loved that!
```
> q <- new_quosure(env(a = iris), "")
> env_print(q@env)
<environment: 0x7f8d7b0af698> [L]
Parent: <environment: package:teal.code>
Bindings:
• a: <df[,5]>
> q
A quosure object containing: a
```

R/quosure-join.R
```diff
@@ -0,0 +1,115 @@
...
+#' - indices of the shared code are not consecutive or don't start from 1
+#' @param x (`Quosure`)
+#' @param y (`Quosure`)
+#' @return `TRUE` if able to join or `character` used to print error message.
+#' @keywords internal
+check_joinable <- function(x, y) {
+  checkmate::assert_class(x, "Quosure")
+  checkmate::assert_class(y, "Quosure")
+
+  common_names <- intersect(ls(x@env), ls(y@env))
```
pawelru, line 77: I have just read that converting symbols to strings might be buggy on different encoding systems (see [here](https://rlang.r-lib.org/reference/env_names.html#encoding)). `rlang::env_names` seems to solve it by forcing to UTF-8

R/quosure-join.R
```diff
@@ -0,0 +1,115 @@
...
+  is_overwritten <- vapply(common_names, function(el) {
+    !identical(get(el, x@env), get(el, y@env))
+  }, logical(1))
```
pawelru, line 80: I am a little bit concerned here about `get()` operation as it could evaluate lazy bindings or promises therefore this functionality could have some side effects. I neither have a counter-example nor a potential solution. Just note to myself...

R/quosure-join.R
```diff
@@ -0,0 +1,115 @@
...
+
+#' If two `Quosure` can be joined
+#'
+#' Checks if two `Quosure` objects can be combined.
+#' They can't be combined if (and):
+#' - both share the same code (identified by `id`)
+#' - indices of the shared code are not consecutive or don't start from 1
+#' @param x (`Quosure`)
+#' @param y (`Quosure`)
+#' @return `TRUE` if able to join or `character` used to print error message.
```
pawelru, line 71: it's always good to assure that function always returns the same data type. This would prevent you doing `expect_true(foo(...))` and `expect_match(foo(...), "...")` at the same time.

This is not the first time where we are facing the problem of returning a character message if something else TRUE. I suggest to always return boolean with additional attribute. Alternatively you might want to use [`checkmate` assertion collections](https://mllg.github.io/checkmate/reference/AssertCollection.html).

R/quosure-join.R
pawelru: Yes you are right. It seems that better naming / docs / ??? is needed.
At the very beginning I was thinking that this is an instance method performing join into itself, i.e. q1.join(q2)
This is however yet another instance constructor basing on two instances, i.e. q_new <- join(q1, q2).
If this should be a constructor - can we introduce `x_new <- new_quosure(...); ...` logic somehow to make it more clear that the new instance is being created.

R/quosure-join.R
```diff
@@ -0,0 +1,115 @@
+#' Join two `Quosure` objects
+#'
+#' Combine two `Quosure` object by merging their environments and the code.
+#' Not every `Quosure` objects can be combined:
+#' - if their environments contains objects of the same name but not identical
+#' - if `y` has unique code element placed before common element. This means that `y`
+#' in the environment of the `y` was evaluated some extra code before which can influence
+#' reproducibility
+#' - more cases to be done
```
pawelru, line 9: make a reference to `check_joinable` and maintain good docs in one place only

R/quosure-class.R
```diff
@@ -0,0 +1,29 @@
...
+#' @rdname Quosure-class
+#' @slot code (`character`) to reproduce the environment
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#'  of the `code` slot.
+#' @slot id (`integer`) random identifier of the code element to make sure uniqueness
+#'  when joining.
+#' @keywords internal
+setClass(
+  "Quosure",
+  slots = c(env = "environment", code = "character", id = "integer"),
```
pawelru, line 14: ID should be an attribute of each element of `code` vector. This would save you assertions of the length between those two and possibly more.

pawelru, line 14: As highlighted before, I don't like `code` being a character field as there are more suitable data structures to keep the _code_. And yes, I am aware that this means no support for comments and I am OK with that.

R/quosure-utils.R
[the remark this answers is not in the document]
pawelru: `rlang::env_clone(env, parent = env_parent(global_env()))`

R/quosure-eval_code.R
pawelru: same reply - I need `deep_clone` call to make it clear that we are initiated new object.

R/quosure-constructor.R
gogonzo: new_quosure.list will be removed soon 👍

pawelru: ok - closing this thread then

R/quosure-utils.R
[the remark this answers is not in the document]
gogonzo: marking as resolved 
- PR here: https://github.com/insightsengineering/teal.code/pull/68

R/quosure-utils.R
gogonzo: marking as resolved 
- PR here: https://github.com/insightsengineering/teal.code/pull/68

R/quosure-eval_code.R
gogonzo: marking as resolved 
- PR here: https://github.com/insightsengineering/teal.code/pull/68

R/quosure-join.R
gogonzo: marking as resolved 
- PR here: https://github.com/insightsengineering/teal.code/pull/68

R/quosure-class.R
gogonzo: marking as resolved 
- PR here: https://github.com/insightsengineering/teal.code/pull/68

R/quosure-constructor.R
```diff
@@ -0,0 +1,103 @@
...
+  "new_quosure",
+  signature = c(env = "environment", code = "character"),
+  function(env, code) {
+    checkmate::assert_character(code, any.missing = FALSE)
+    new_env <- .copy_env(env)
+    lockEnvironment(new_env)
+
+    if (is.null(names(code)) && length(code)) {
+      names(code) <- make.unique(rep("initial code", length(code)))
+    }
```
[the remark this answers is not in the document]
gogonzo: marking as resolved 
- PR here: https://github.com/insightsengineering/teal.code/pull/68

R/quosure-constructor.R
gogonzo: Marking as resolved:
discussed yesterday. We will make `data` as a new class and this data object will have wrapper get_code 
which will be used in the `new_quosure(list2env(data), get_code(data))`

https://github.com/insightsengineering/teal/issues/731#issue-1374679782

R/quosure-constructor.R
gogonzo: Marking as resolved:
discussed yesterday. We will make `data` as a new class and this data object will have wrapper get_code 
which will be used in the `new_quosure(list2env(data), get_code(data))`

https://github.com/insightsengineering/teal/issues/731#issue-1374679782

R/quosure-join.R
gogonzo: marking as resolved:
I guess this should be clear now Marking as resolved. Please comment the PR if something require improvement 

https://github.com/insightsengineering/teal.code/pull/68/files#diff-3f9293e71b2b2e30dc82142853ab25338b5b3f4a594fe1f326c932a64083aaba

R/quosure-show.R
gogonzo: marking as resolved

PR here: #68

R/quosure-join.R
gogonzo: check_joinable is a private utility. Is it ok to expose this function in the public docs?

R/quosure-join.R
gogonzo: marking as resolved:

https://github.com/insightsengineering/teal.code/pull/68/files#diff-3f9293e71b2b2e30dc82142853ab25338b5b3f4a594fe1f326c932a64083aaba

R/quosure-join.R
pawelru: You can still have a roxygen documentation for private functions and manage public/private in the pkgdown config file.

R/quosure-constructor.R
```diff
@@ -0,0 +1,103 @@
...
+
+#' @rdname new_quosure
+#' @export
+setMethod(
+  "new_quosure",
+  signature = c(env = "environment", code = "character"),
+  function(env, code) {
+    checkmate::assert_character(code, any.missing = FALSE)
+    new_env <- .copy_env(env)
+    lockEnvironment(new_env)
```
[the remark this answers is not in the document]
gogonzo: resolve conversation:
addressed here
https://github.com/insightsengineering/teal.code/pull/68/files

R/quosure-join.R
gogonzo: Yes we can, but the last time we discussed this we agreed that:
1. all private functions should have @keyword internal
2. private functions shouldn't be exposed in the public docs (to avoid users using private functions)
3. internal functions not listed in pkgdown

We can make a reference in the docs, but that's against (2). I'm totally against listing .check_joinable in the pkgdown

R/quosure-class.R
gogonzo: we can do this but please know that you can't assign attribute to a single element of the expression because it's like a vector.


```r
a <- as.expression(quote(a <- 1))
attr(a, "id") <- sample.int(.Machine$integer.max, size = 1)

b <- as.expression(quote(b <- 2)) 
attr(b, "id") <- sample.int(.Machine$integer.max, size = 1)

code <- c(a, b)
attr(code[1], "id")
attr(code[2], "id")
```

we will need to do something like this:

```r
code <- as.expression(quote(a <- 1))
attr(code, "id") <- sample.int(.Machine$integer.max, size = 1)

new_code <- as.expression(quote(b <- 2)) 
new_attrs <- c(attr(code, "id"), sample.int(.Machine$integer.max, size = 1))
code <- c(code, new_code)
attr(code, "id") <- new_attrs
```

If you feel that this is important please create the issue in teal.code

pawelru: How about the following?
```
a <- as.expression(quote(a <- 1))
attr(a, "id") <- sample.int(.Machine$integer.max, size = 1)

b <- as.expression(quote(b <- 2)) 
attr(b, "id") <- sample.int(.Machine$integer.max, size = 1)

code <- list(a, b)
attr(code[[1]], "id")
attr(code[[2]], "id")
```

or alternatively:

```
a <- setNames(list(as.expression(quote(a <- 1))), sample.int(.Machine$integer.max, size = 1))
b <- setNames(list(as.expression(quote(b <- 2))), sample.int(.Machine$integer.max, size = 1))
c(a, b)
```

gogonzo: 1. is not enough because you can provide code as n-element expression 
```r
eval_code(<qenv>, as.expression(quote(a <- 1), quote(b <- 2)))
```
There will be some additional code to loop through each and change from vector to a list

2. Look how funny
```r
a <- c("a", "b", "c")
names(a) <- c(3, 2, 1)
a[1]
#  3 
# "a" 
```

R/quosure-join.R
```diff
@@ -0,0 +1,115 @@
...
+  } else if (!identical(shared_in_x, shared_in_y)) {
+    paste(
+      "The common code of joined objects doesn't have the same indices. It means that `x` and `y`",
+      "can't be joined together as it's impossible to determine the evaluation's order.",
+      collapse = ""
+    )
+  } else {
+    paste(
+      "The common code of joined object does not start from index = 1.",
+      "It means that joined object(x) has some extra code elements before.",
```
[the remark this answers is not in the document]
pawelru: https://github.com/insightsengineering/teal.code/issues/69

R/quosure-class.R
pawelru: new issue: https://github.com/insightsengineering/teal.code/issues/70

R/quosure-join.R
pawelru: new issue https://github.com/insightsengineering/teal.code/issues/71

R/quosure-join.R
asbates: I agree. As a user it can be annoying when a function returns 2 types of objects. ('user' includes developer w/in package)

vignettes/qenv.Rmd
```diff
@@ -0,0 +1,144 @@
...
+      data_q,
+      bquote(p <- hist(iris_data[, .(input$option)]))
+    )
+  })
+
+  # display output
+  output$plot <- renderPlot(output_q()[["p"]])
+
+  # code is displayed when "Show Code" button is clicked on
+  verbatim_popup_srv("rcode", reactive(paste(get_code(output_q()), collapse = "\n")), title = "Code")
```
nikolas-burkoff, line 133: @gogonzo do we actually need the paste and collapse here?
```suggestion
  verbatim_popup_srv("rcode", reactive(get_code(output_q())), title = "Code")
```

gogonzo: I think it's needed for renderText -> verbatimTextOutput. I'll check it

R/quosure-join.R
gogonzo: This function is internal, user shouldn't use this function at all. `.check_joinable` is used only in the `join` in following way:
```r
  join_validation <- .check_joinable(x, y)

  # join expressions
  if (!isTRUE(join_validation)) {
    stop(join_validation)
  }
```

Alternatively, we can move `stop` to the .check_joinable and forget the whole thing

vignettes/teal-code.Rmd
```diff
@@ -9,51 +9,15 @@ vignette: >
   %\VignetteEncoding{UTF-8}
 ---
 
+NOTE: Chunks are being deprecated `qenv` objects should be used instead - new vignettes are on their way.
```
nikolas-burkoff, line 12:
```suggestion

```

vignettes/teal-code.Rmd
```diff
@@ -9,51 +9,15 @@ vignette: >
...
-The code chunks were introduced to allow the development of reproducible `shiny`/`teal` modules. A reproducible module includes a "Show R Code" button which can display the code needed to reproduce the outputs of the module. Inside the modal, which pops up after clicking the button, you should see:
-
-1. The code loading the necessary libraries - by `teal::get_code`
-2. The code used for loading and filtering data sets  - by `teal::FilteredData`
-3. The code used for merging multiple data sets - by `teal.transform::data_merge_srv`
-4. The code generating the outputs - by **code chunks**
-
-
-The code chunks were especially designed for the code leading from an input data set to the outputs. To get more information about the concept of chunks, please refer to article [`basic chunks`](https://insightsengineering.github.io/teal.code/articles/basic_chunks.html).
+For further details see the [`qenv` vignette](https://insightsengineering.github.io/teal.code/articles/qenv.html).
```
nikolas-burkoff, line 23:
```suggestion
For further details see the [`qenv` vignette](https://insightsengineering.github.io/teal.code/articles/qenv.html).

Note: the older method of handling reproducibility, `chunks`, have now been deprecated and `qenv` should be used instead.


```

vignettes/basic_chunks.Rmd
```diff
@@ -1,251 +1,13 @@
 ---
 title: "Basic chunks"
 author: "NEST coreDev"
-date: "2022-04-22"
+date: "2022-10-01"
```
nikolas-burkoff, line 4:
```suggestion
date: "2022-11-03"
```

vignettes/advanced_chunks.Rmd
```diff
@@ -1,402 +1,14 @@
 ---
 title: "Advanced chunks"
 author: "NEST coreDev"
-date: "2022-05-09"
+date: "2022-10-01"
```
nikolas-burkoff, line 4:
```suggestion
date: "2022-11-03"
```

R/qenv-show.R
```diff
@@ -0,0 +1,18 @@
...
+#'
+#' Prints the `qenv` object
+#' @param object (`qenv`)
+#' @return nothing
+#' @importFrom methods show
+#' @examples
+#' q1 <- new_qenv(code = "print('a')", env = new.env())
+#' q1
+#' @export
+setMethod("show", "qenv", function(object) {
```
nikolas-burkoff, line 11: See https://github.com/insightsengineering/teal.code/issues/88 (for later)

vignettes/qenv.Rmd
```diff
@@ -0,0 +1,144 @@
+---
+title: "`qenv`"
+author: "NEST coreDev"
+date: "2022-10-01"
```
nikolas-burkoff, line 4:
```suggestion
date: "2022-11-03"
```

vignettes/qenv.Rmd
nikolas-burkoff: Seems to work without it -> verbatim popup must be smart enough

tests/testthat/test-qenv_constructor.R
```diff
@@ -0,0 +1,82 @@
...
+  q3 <- new_qenv(code = quote(a <- 1), env = list2env(list(a = 1)))
+  testthat::expect_identical(parent.env(q1@env), parent.env(.GlobalEnv))
+  testthat::expect_identical(parent.env(q1@env), parent.env(q2@env))
+  testthat::expect_identical(parent.env(q1@env), parent.env(q3@env))
+})
+
+testthat::test_that("provided environment is copied so the change of the environment won't affect qenv", {
+  a <- 1L
+  q1 <- new_qenv(quote(a <- 1), env = environment())
+  b <- 2L
```
nikolas-burkoff, line 24: should this be @gogonzo 
```suggestion
  a <- 2L
```

tests/testthat/test-qenv_eval_code.R
```diff
@@ -0,0 +1,154 @@
...
+
+testthat::test_that("library have to be called separately before using function from package", {
+  q1 <- eval_code(new_qenv(), quote(library(checkmate)))
+  # library call adds package env in the search path just over .GlobalEnv
+  #   it means that q3@env before the call was a parent of .GlobalEnv but not after the call
+
+  q2 <- eval_code(q1, quote(assert_number(1)))
+  testthat::expect_identical(parent.env(q2@env), parent.env(.GlobalEnv))
+
+  detach("package:checkmate", unload = TRUE)
```
nikolas-burkoff, line 35: Just confirming this test leaves the status of the checkmate package unchanged (we don't need a on.exit(detach("package:checkmate")) on line 29 to make sure that it is detached if something goes wrong in lines 32 and 33? @gogonzo

nikolas-burkoff (approved):
Lets do this 🦾

DESCRIPTION
```diff
@@ -19,14 +19,17 @@ Depends:
     R (>= 4.0)
 Imports:
     checkmate,
+    cli,
```
pawelru, line 22: can we remove `cli` dependency? It's a big package

nikolas-burkoff: if we do then we don't fix this:
https://github.com/insightsengineering/teal.code/issues/86 - unless you want to fix it a different way

pawelru:
And we are one step away from test coverage perfection :). In the covr report there is only one line non-covered in tests:
```
Filename                 Stmts    Miss  Cover    Missing
---------------------  -------  ------  -------  ---------------------------------------------------------------------------------------------------------------------------------------------
R/qenv-eval_code.R          45       1  97.78%   89
```

DESCRIPTION
pawelru: that's my question - can we do (name that thing) differently not involving cli package?

nikolas-burkoff: so this is what we call:

```
> cli::ansi_strip
function (string, sgr = TRUE, csi = TRUE, link = TRUE) 
{
    if (!is.character(string)) 
        string <- as.character(string)
    string <- enc2utf8(string)
    stopifnot(is_flag(sgr), is_flag(csi), is_flag(link))
    clean <- .Call(clic_ansi_strip, string, sgr, csi, link)
    class(clean) <- setdiff(class(clean), c("cli_ansi_string", 
        "ansi_string"))
    clean
}
<bytecode: 0x0000018dd338c3e0>
<environment: namespace:cli>
```

nikolas-burkoff: and dplyr imports cli so in practice ppl using this package will have cli installed but if there's an easy way round this then happy to change (or put issue in bnacklog to change)

nikolas-burkoff: we could add it into suggests like rlang and only call it if installed?

pawelru: Here I am more concerned about the package size and the installation process. Up to you how would you like to process it.

pawelru: > we could add it into suggests like rlang and only call it if installed?

This only makes sense if a particular piece of code is unlikely to be called assuming _standard_ workflow. E.g. a _wide_ tree-based workflow / case when where there are many many possibilities. Here, if I understand everything correctly it's a part of standard logic so it's very likely that it will be executed.

nikolas-burkoff: so I assume that these funny terminal printing characters added by rlang are only added if cli is installed (cli is a suggests in rlang only) - therefore I think it makes sense for us to use suggests as well - as striping these terminal printing characters out only makes sense if cli (via rlang) added them in the first place

nikolas-burkoff: so rlang is only in the suggests but needs to be imported - whoops

gogonzo:
For anyone concerned about deep coping the objects. There are no deep copies in `eval_code` and `join`, we only create a new environment each time, set parent to .globalEnv and we create a new reference to the same pointer in the new env.

```r
library(teal.code)
library(lobstr)
x <- data.frame(a = runif(10))

q1 <- new_qenv(env = list2env(list(x = x)), code = "x <- data.frame(a = 1)")

identical(lobstr::ref(q1[["x"]]), lobstr::ref(x))
q2 <- eval_code(q1, quote(x$b <- runif(10)))

!identical(lobstr::ref(q2[["x"]]), lobstr::ref(q1[["x"]]))
identical(lobstr::ref(q2[["x"]]$a), lobstr::ref(q1[["x"]]$a))

q3 <- eval_code(q2, quote(y <- x))
identical(q3[["y"]], q2[["x"]])

q4 <- join(q2, q3)
identical(lobstr::ref(q4[["y"]]), lobstr::ref(q2[["x"]]))
```

donyunardi:
`qenv` vignette was clear and easy to follow. Great job!

PR #68, opened 2022-09-16
@code character to expression 

### LINKED ISSUE #67: qenv to store code as expression
Followup to the @pawelru comments:
- https://github.com/insightsengineering/teal.code/pull/66#discussion_r972164927
- https://github.com/insightsengineering/teal.code/pull/66#discussion_r972236785

Since qenv returns ´quosure.error´ when failing, debug info module doesn't make sense so we don't need to store @code as a named character but as a expression. What is needed:
- remove eval_code.character method and move eval to the eval_code.expression
- remove new_quosure.character
- change all occurrences where 'eval_code' uses code as character
- remove all ´name´ in the eval_code
- change get_code to ´function(object, deparse = TRUE)´ - should still by default return the code as character with option ´deparse´ to optionally return expression

gogonzo:
closes #67 

Also needs
- tmg https://github.com/insightsengineering/teal.modules.general/pull/461
- tmc https://github.com/insightsengineering/teal.modules.clinical/pull/600
- t.o https://github.com/insightsengineering/teal.osprey/pull/179
- t.g https://github.com/insightsengineering/teal.goshawk/pull/189


R/quosure-constructor.R
```diff
@@ -45,8 +33,10 @@ setMethod(
   "new_quosure",
   signature = c(env = "environment", code = "expression"),
   function(env, code) {
-    code_char <- as.character(code)
-    new_quosure(env = env, code = code_char)
+    new_env <- rlang::env_clone(env, parent = parent.env(.GlobalEnv))
```
gogonzo, line 36: instead .copy_env rlang::env_clone


https://github.com/insightsengineering/teal.code/pull/66#discussion_r972181884

R/quosure-constructor.R
```diff
@@ -45,8 +33,10 @@ setMethod(
   "new_quosure",
   signature = c(env = "environment", code = "expression"),
```
pawelru, line 34: please put this method first as this is the "main" constructor that all the others are using

R/quosure-constructor.R
```diff
@@ -9,12 +9,11 @@
 #' @param env (`environment`) Environment being a result of the `code` evaluation.
 #'
 #' @examples
-#' new_quosure(env = list2env(list(a = 1)), code = "a <- 1")
```
pawelru, line 12: this should stay as we are still supporting chars but change the implementation only
you can actually reorder to showoff that expression is preferred compared to char

R/quosure-eval_code.R
```diff
@@ -8,37 +8,35 @@
...
+
+#' @rdname eval_code
+#' @export
+setMethod("eval_code", signature = c("Quosure", "expression"), function(object, code) {
   id <- sample.int(.Machine$integer.max, size = length(code))
 
   object@id <- c(object@id, id)
-  object@code <- .keep_code_name_unique(object@code, code)
+  object@env <- rlang::env_clone(object@env, parent = parent.env(.GlobalEnv))
+  object@code <- c(object@code, code)
```
pawelru, line 35: so here I would love to have the following:
```
new_object <- clone(object) # new method in which we are doing rlang::env_clone etc.
new_object@id <- c(new_object@id, id)
new_object@code <- c(...)
eval(...)
```

This new (private) method is then used here in eval and also in join. It makes it more explicit that new object is being created and then further modified

What's your view on this?

R/quosure-get_code.R
```diff
@@ -2,16 +2,16 @@
 #'
 #' @name get_code
 #' @param object (`Quosure`)
+#' @param deparse (`logical(1)`) if the returned code should be converted to character.
```
pawelru, line 5: should this be handled on the caller side?

R/quosure-eval_code.R
```diff
@@ -8,37 +8,35 @@
...
-    code <- paste(code, collapse = "\n")
-    names(code) <- name
-  }
+setMethod("eval_code", signature = c("Quosure", "character"), function(object, code) {
+  eval_code(object, code = parse(text = code, keep.source = FALSE))
+})
+
+#' @rdname eval_code
+#' @export
+setMethod("eval_code", signature = c("Quosure", "expression"), function(object, code) {
```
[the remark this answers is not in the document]
nikolas-burkoff: done

R/quosure-get_code.R
nikolas-burkoff: That will then mean every module (and module developer) will need to change from:


```
...
verbatim_content = reactive(teal.code::get_code(output_q())),
...

card$append_src(paste(teal.code::get_code(output_q()), collapse = "\n"))
```

to adding extra deparse statements - which I'm not sure is what we want?

gogonzo: We can remove `deparse` and limit `get_code` to always return character. I don't see the sense of returning "raw" `@code` (expression).

R/quosure-join.R
```diff
@@ -39,12 +39,11 @@ setMethod("join", signature = c("Quosure", "Quosure"), function(x, y) {
 
   id_unique <- !y@id %in% x@id
   x@id <- c(x@id, y@id[id_unique])
-  x@code <- .keep_code_name_unique(x@code, y@code[id_unique])
+  x@code <- c(x@code, y@code[id_unique])
 
   # insert (and overwrite) objects from env2 to env
```
pawelru, line 44:
```suggestion
  # insert (and overwrite) objects from y to x
```

R/quosure-eval_code.R
gogonzo: I assume that `clone` wouldn't clone the whole `object` but only the environment. `object` is cloned automatically by R internals when modified. Sounds too much for me

pawelru: ok let's leave it then

pawelru (approved):

Issue #70, opened 2022-09-16
[Suggestion] Modifying the slot design for the qenv class


pawelru:
ID should be an attribute of each element of `code` vector. This would save you assertions of the length between those two and possibly more.

_Originally posted by @pawelru in https://github.com/insightsengineering/teal.code/pull/66#discussion_r972163426_

donyunardi:
The class' slots has grown during development and with all the works that has been going in to the feature branch, maybe it'd be best if we reserve this for the next increment.

The team has reviewed this and assigned story point but still need to decide if this is necessary to do.

gogonzo:
same applies to:
- warnings
- messages

PR #79, opened 2022-09-27
qenv: warnings and messages

### LINKED ISSUE #63: Handle warnings and messages in qenv
https://github.com/insightsengineering/teal.code/pull/61 handles errors within chunks but not warnings or messages (nor outputs like print statements)

1) Store warnings and messages in `qenv` slots or maybe as atrtributes on the code slot items- see how this is done in chunks [here](https://github.com/insightsengineering/teal.code/blob/main/R/chunk.R#L98-L119) - the errors have been handled already?
2) Add `get_warning()` (and get_message() - though that probably won't be used) functions for `qenv` (make sure they work with `qenv.error` as well - could just return NULL or some error?) which could return NULL if there are no warnings
3) Have a `uiOutput` in the modules underneath/next to the show R code button
4) In renderUI have a `teal.widgets::verbatim_popup_ui` with say `req(get_warning(output_q()))` to make sure it's only shown if there's warnings - and clicking on the button shows the warnings
5) Make changes in teal.modules.general, teal.modules.clinical, teal.osprey, teal.gohawk and possibly 1 or 2 of teal's vignettes

Optionally: 6) If there are warnings created then display a notification when they are created  `shiny::showNotification`

Also update vignette in teal.code showing this functionality

Also add tests in teal.code for this functionality

Branch off teal_refactor@main onto x@teal_refactor@main for some x (same branch for all PRs)

mhallal1:
closes #63 

1) Every time the code is evaluated in `qenv`, if a warning or message is detected, they are added as attributes to the `code`.
2) Methods to `get_warnings` and `get_messages` are added.
3) Modules will have to be updated as in this branch: https://github.com/insightsengineering/teal.modules.general/pull/465

Please test with the app below, add warnings and messages as required, test also with tmg, tmc... modules:
```
library(shiny)
library(teal.widgets)
library(magrittr)
library(teal.code)

# create an initial qenv with the data in
data_q <- new_qenv() %>% eval_code("iris_data <- iris")
ui <- fluidPage(
  shinyjs::useShinyjs(),
  radioButtons(
    "option", "Choose a column to plot (Petal.Width give warning):",
    c("Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width", "error_option")
  ),
  # button for the Code to be shown when clicked on
  verbatim_popup_ui("warning", button_label = "Show Warning"),
  verbatim_popup_ui("rcode", button_label = "Show Code"),
  plotOutput("plot")
)
server <- function(input, output, session) {
  # create a qenv containing the reproducible output
  output_q <- reactive({
    req(input$option)
    if (input$option == "Petal.Width") {
      eval_code(
        data_q,
        bquote(p <- hist(iris_data[, .(input$option)], non_arg = ""))
      )
    } else {
      eval_code(
        data_q,
        bquote(p <- hist(iris_data[, .(input$option)]))
      )
    }
  })
  # display output
  output$plot <- renderPlot(output_q()[["p"]])
  # code is displayed when "Show Code" button is clicked on
  verbatim_popup_srv(
    "warning",
    reactive(paste(get_warnings(output_q()), collapse = "\n")),
    title = "Warning",
    disabled = reactive(is.null(get_warnings(output_q())))
  )
  verbatim_popup_srv("rcode", reactive(paste(get_code(output_q()), collapse = "\n")), title = "Code")
}
shinyApp(ui, server)

```

related PRs:
-teal.modules.general: https://github.com/insightsengineering/teal.modules.general/pull/465
-teal.modules.clinica: https://github.com/insightsengineering/teal.modules.clinical/pull/621
-teal.goshawk: https://github.com/insightsengineering/teal.goshawk/pull/191
-teal.osprey: https://github.com/insightsengineering/teal.osprey/pull/181
-teal: https://github.com/insightsengineering/teal/pull/739


R/qenv-eval_code.R
```diff
@@ -44,6 +44,18 @@ setMethod("eval_code", signature = c("qenv", "expression"), function(object, cod
         class = c("qenv.error", "try-error", "simpleError"),
         trace = object@code
       )
+    },
+    warning = function(w) {
```
nikolas-burkoff, line 48: So I think you need to use the withCallingHandlers that chunks used to catch warnings but then continue evaluating the code

R/qenv-eval_code.R
```diff
@@ -44,6 +44,18 @@ setMethod("eval_code", signature = c("qenv", "expression"), function(object, cod
...
+      attr(object@code, "warning") <- as.character(w)
```
nikolas-burkoff, line 49: So `object@code` is a vector - we could do `attr(object@code[length(object@code])` to add the warning onto the specific code (and we should do `attr(object@code[length(object@code], "warning") <- c(.., as.character(w)) to catch multiple warnings).

If we're just adding it to the code as a whole then I would actually have object@warnings and store them there instead

R/get_warnings.R
```diff
@@ -0,0 +1,33 @@
...
+#' @examples
+#' data_q <- new_qenv() %>% eval_code("iris_data <- iris")
+#' warning_qenv <- eval_code(
+#'   data_q,
+#'   bquote(p <- hist(iris_data[, .("Sepal.Length")], ff = ""))
+#' )
+#' get_warnings(warning_qenv)
+#'
+get_warnings <- function(object) {
+  attr(object@code, "warning")
```
nikolas-burkoff, line 17: we should also make sure joining qenv's together transfers the warning and messages as well

R/qenv-eval_code.R
mhallal1: added.

R/get_warnings.R
```diff
@@ -0,0 +1,33 @@
...
+#'
+#' @examples
+#' data_q <- new_qenv() %>% eval_code("iris_data <- iris")
+#' warning_qenv <- eval_code(
+#'   data_q,
+#'   bquote(p <- hist(iris_data[, .("Sepal.Length")], ff = ""))
+#' )
+#' get_warnings(warning_qenv)
+#'
+get_warnings <- function(object) {
```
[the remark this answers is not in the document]
mhallal1: done.

R/qenv-eval_code.R
nikolas-burkoff: done

R/get_warnings.R
nikolas-burkoff: done

R/qenv-get_warnings.R
```diff
@@ -0,0 +1,41 @@
+#' Get the warnings of qenv object
+#'
+#' @param object (`qenv`)
+#'
+#' @return `character` containing warning information or `NULL` if no warnings
+#' @export
+#'
+#' @examples
+#' data_q <- new_qenv() %>% eval_code("iris_data <- iris")
```
nikolas-burkoff, line 9: Need library(magrittr) for the example and add magrittr to suggests or just do

data_q <- eval_code(new_qenv(), "iris_data <- iris")

R/qenv-eval_code.R
```diff
@@ -28,24 +28,45 @@ setMethod("eval_code", signature = c("qenv", "expression"), function(object, cod
...
+            conditionMessage(e),
+            paste(code, collapse = "\n ")
+          ),
+          class = c("qenv.error", "try-error", "simpleError"),
+          trace = object@code
+        )
+      }
+    ),
+    warning = function(w) {
+      current_warnings <<- paste(current_warnings, suppressWarnings(as.character(w)))
```
nikolas-burkoff, line 54: As we have slots for warnings and messages we may be able to replace:

`current_warnings <- ""` with `object@warnings <- c(object@warnings, "")` and then

`current_warnings <<- paste(current_warnings, suppressWarnings(as.character(w)))` with

`object@warnings[length(object@warnings)] <- paste(object@warnings[length(object@warnings)] , suppressWarnings(...)`

and then remove lines 66 and 67? As we have tests we can easily check

tests/testthat/test-qenv_constructor.R
```diff
@@ -4,6 +4,8 @@ testthat::test_that("constructor returns qenv if nothing is specified", {
   testthat::expect_identical(ls(q@env), character(0))
   testthat::expect_identical(q@code, expression())
   testthat::expect_identical(q@id, integer(0))
+  testthat::expect_identical(q@warnings, character(0))
+  testthat::expect_identical(q@messages, character(0))
```
nikolas-burkoff, line 8: We should probably test that if you create a qenv using code = "..." i.e. line 11 then you can also have the correct number of elements in q@warnings (i.e. testing the rep("", length(code)) line in the constructor

nikolas-burkoff (commented):
Looking good - a couple of minor tweaks but I think this is pretty much done

R/qenv-get_warnings.R
mhallal1: done.

tests/testthat/test-qenv_get_warnings.R
```diff
@@ -0,0 +1,44 @@
...
+  q <- new_qenv() %>% eval_code(bquote(error("This is a error!")))
+  testthat::expect_null(get_warnings(q))
+})
+
+testthat::test_that("get_warnings accepts a qenv object with no warning and returns NULL", {
+  q <- new_qenv() %>% eval_code(bquote("x <- 1"))
+  testthat::expect_null(get_warnings(q))
+})
+
+testthat::test_that("get_warnings accepts a qenv object with 2 warnings", {
```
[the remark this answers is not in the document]
mhallal1: done.

R/qenv-eval_code.R
mhallal1: We go back to the issue where object@warnings  keeps re initializing and rendering empty warnings slot :(

tests/testthat/test-qenv_eval_code.R
```diff
@@ -108,3 +108,47 @@ testthat::test_that("an error when calling eval_code returns a qenv.error object
...
+  testthat::expect_s4_class(q, "qenv")
+  testthat::expect_equal(
+    q@warnings,
+    c(
+      "",
+      paste0(
+        "\"ff\" is not a graphical parameter",
+        "\"ff\" is not a graphical parameter",
+        "\"ff\" is not a graphical parameter",
+        "\"ff\" is not a graphical parameter"
```
asbates, line 124: I was confused on an initial read of this test as to why there are 4 of the same warning message. I wonder if a simpler example would work here? Maybe something like `q <- eval_code(new_qenv(), quote("warning('warning')"))`.

asbates (commented):
I would consider adding a section to the `qenv` vignette. Errors are briefly mentioned but not warnings or messages.

tests/testthat/test-qenv_eval_code.R
mhallal1: @asbates I would leave it as it is now to make sure that any future change in the code still reports all the warnings rendered and not only a specific one.
For instance, this code:
```
iris_data <- iris
p <- hist(iris_data[, 'Sepal.Length'], ff = '')
```
renders four warnings:
```
Warning messages:
1: In plot.window(xlim, ylim, "", ...) : "ff" is not a graphical parameter
2: In title(main = main, sub = sub, xlab = xlab, ylab = ylab, ...) :
  "ff" is not a graphical parameter
3: In axis(1, ...) : "ff" is not a graphical parameter
4: In axis(2, ...) : "ff" is not a graphical parameter
```
and we optimally have to report the 4 of them to the user to not truncate any information.
Does this make sense to you @asbates ?

nikolas-burkoff:
> I would consider adding a section to the qenv vignette. Errors are briefly mentioned but not warnings or messages.

Yup that makes sense (the qenv vignette was not merged into teal_refactor@main when this branch was being worked on)

@asbatescan you add that to this branch?

tests/testthat/test-qenv_eval_code.R
nikolas-burkoff: @asbates can you update this? thanks

asbates: @mhallal1 @nikolas-burkoff It makes more sense to me now. This is a test to ensure that multiple warnings from a single line of code are captured. What about updating the test description instead of a comment/another code line? Something that invokes the test purpose a little more clearly. I can take care of this.

asbates:
> > I would consider adding a section to the qenv vignette. Errors are briefly mentioned but not warnings or messages.
> 
> Yup that makes sense (the qenv vignette was not merged into teal_refactor@main when this branch was being worked on)
> 
> @asbates can you add that to this branch?

Will do.

gogonzo (commented):
I'm not sure whether we want to keep warnings and messages and show them to users. 
There is consensus about eval_details that we don't have to provide a debugging system for qenv except handling errors which break the app (@pawelru can confirm). 
If we want to keep warnings I'd like to not expose them in the constructor. Where to keep them then?
1. code can be a `list` instead of `expression` and each element can have the attribute (warning, message, and possibly more). We would not change the constructor.
2. Storing warnings in the `baseenv()[["last.warning"]]` (same as normal warnings). We can achieve this in this way in eval_code.
```r
tryCatch(....,
  warning = function(w) {
    w[["call"]] <- code # we need to convert expression to call
    warning(w)
    return(object)
  }
))
```
In above warnings will be appended to the warnings container (together with other warnings nor related to qenv execution)

nikolas-burkoff:
> I'm not sure whether we want to keep warnings and messages and show them to users.

@kumamiao - I believe we do need a way for app users to see warnings

nikolas-burkoff:
When you say 
>expose them in the constructor.

What exactly do you mean?

gogonzo:
> What exactly do you mean? I assume having it in methods::new (as it's not exposed in new_qenv)
Sorry, I was wrong with the exposition as `new_qenv` doesn't have this as parameter but they (warnings and msgs) are only in the slots. This is better than I initially thought.

> @kumamiao - I believe we do need a way for app users to see warnings
Let's continue discussion about this. In what cases we would like these warnings/messages to be shown to users, and if users have to know which R call caused this.

nikolas-burkoff:
> 1. code can be a list instead of expression and each element can have the attribute (warning, message, and possibly more). We would not change the constructor.

In that case I'd argue why not use S3 - if code is a list we then need to add validation to it otherwise it could be a list of anything :)  But yes it could be done and could be quite nice you could also have the id attribute as well for https://github.com/insightsengineering/teal.code/issues/70

> 2. Storing warnings in the baseenv()[["last.warning"]] (same as normal warnings). We can achieve this in this way in eval_code.

I think we need to separate the warnings from the qenv from the warnings of the app as the former matter for generation of the output and users may care about but the latter are just noise for them

kpagacz:
> I think we need to separate the warnings from the qenv from the warnings of the app as the former matter for generation of the output and users may care about but the latter are just noise for them

Here are a couple of ideas on what to do with warnings and messages from the application/SRC code:
* brutally piped them to a queue created for that purpose in `teal` (or another package that makes sense, or a new package) and retrieve them when needed. Might have the added benefit of being pan-`teal` in its nature - probably most of the packages could pipe application/SRC code errors there. And the queue probably needs not to be completely stupid - a limit on the number of elements is probably needed.
* defer to an already existing package: [rlang](https://rlang.r-lib.org/reference/abort.html) has a sensible way to handle these. Supports custom errors/warnings/messages ;)

gogonzo:
> > 1. code can be a list instead of expression and each element can have the attribute (warning, message, and possibly more). We would not change the constructor.
> 
> In that case I'd argue why not use S3 - if code is a list we then need to add validation to it otherwise it could be a list of anything :) But yes it could be done and could be quite nice you could also have the id attribute as well for #70
> 

Exactly, having code as the list will address also #70 . `@code` could be a class for validation purposes.

> > 2. Storing warnings in the baseenv()[["last.warning"]] (same as normal warnings). We can achieve this in this way in eval_code.
> 
> I think we need to separate the warnings from the qenv from the warnings of the app as the former matter for generation of the output and users may care about but the latter are just noise for them
> 
> > and if users have to know which R call caused this.
> 
> Because of joining qenvs even if users don't need to know which call caused it, we need to know to make sure the joined quosure makes sense (we can't just concatenate all the warnings from both source qenvs etc.)

Yes, we need a separation maybe `rlang` has some answer to this? @kpagacz points a promising direction. Maybe we can save qenv warnings somewhere else than `baseenv()[["last.warning"]]` - but please be aware that we shouldn't keep the warnings in the package environment as it might be shared across the shiny sessions - we need to figure out a new location.
Which convince me that it should be somewhere in the class.

R/qenv-eval_code.R
```diff
@@ -28,24 +28,46 @@ setMethod("eval_code", signature = c("qenv", "expression"), function(object, cod
...
+        current_warnings <<- paste0(current_warnings, conditionMessage(w))
+        invokeRestart("muffleWarning")
+      },
+      message = function(m) {
+        current_messages <<- paste0(current_messages, conditionMessage(m))
+        invokeRestart("muffleMessage")
+      }
+    )
+    if (!is.null(x)) {
+      return(x)
```
gogonzo, line 63: I would use `tryInvokeRestart` instead of `invokeRestart` as it might fail in some cases.

`if (!is.null(x)) return(x)` is problematic. Meaning that if successful then escape. We need to escape the loop/function only in there is a `qenv.error`. Otherwise, continue evaluation.

We can also use withCallingHandlers only - instead of wrapping tryCatch within.



```suggestion
  for (code_line in code) {
    x <- withCallingHandlers(
      eval(code_line, envir = object@env),
      error = function(e) {
        errorCondition(
          message = sprintf(
            "%s \n when evaluating qenv code:\n %s",
            conditionMessage(e),
            paste(code, collapse = "\n ")
          ),
          class = c("qenv.error", "try-error", "simpleError"),
          trace = object@code
        )
      },
      warning = function(w) {
        current_warnings <<- paste0(current_warnings, conditionMessage(w))
        tryInvokeRestart("muffleWarning")
      },
      message = function(m) {
        current_messages <<- paste0(current_messages, conditionMessage(m))
        tryInvokeRestart("muffleMessage")
      }
    )

    if (inherits(x, "qenv.error")) {
      return(x)
```

gogonzo (commented):
Couple comments. I'm happy with the solution having warnings and messages saved in the class.

gogonzo (approved):
👍

PR #82, opened 2022-10-21
qenv: return NULL when var not found

### LINKED ISSUE #57: Quosure - get operator should return a NULL for nomatch
The Quosure get operator should return a NULL for nomatch, the same as chunks and base R environment / list.

```
ee <- new.env()
ee[["STH"]]
# NULL
ll <- list()
ll[["STH"]]
# NULL
```

mhallal1:
closes #57 

Returns `NULL` when variable not found in `qenv` or with `qenv-error` when calling `get_var` or `[[var]]`.


nikolas-burkoff (dismissed):
So I think that qenv.error's behaviour should not be changed only qenv's @gogonzo ?

PR #91, opened 2022-11-09
Enable checks for `workflow_dispatch` event

walkowif:


.github/workflows/check.yaml
```diff
@@ -25,13 +25,12 @@ jobs:
     secrets:
       REPO_GITHUB_TOKEN: ${{ secrets.REPO_GITHUB_TOKEN }}
   coverage:
-    if: github.event_name == 'pull_request'
-    name: Coverage 📔 
+    name: Coverage 📔
```
arkadiuszbeer, line 28:
```suggestion
    if: github.event_name != 'push'
    name: Coverage 📔
```
What with `if` condition?

arkadiuszbeer (approved):
LGTM

.github/workflows/check.yaml
walkowif: As discussed, let's enable coverage check also for pushes to `main`.

PR #101, opened 2023-01-06
Use rlang::env_print

### LINKED ISSUE #88: Show method for `qenv` use `rlang::env_print`
Currently we print the contents of the environment ourselves - we could use rlang's function to do it for us

nikolas-burkoff:
Closes #88

image


gogonzo (approved):
👍

PR #103, opened 2023-01-27
update the value of `deprecate_warn(where=)`

donyunardi:
The next release version will be 0.3.0.
Updating the value of `deprecate_warn(where=)` in chunks feature.


gogonzo:
@donyunardi in `"deprecate_warn"` there is `when = "0.3.0"` while the version is still 0.2.1. In my opinion this doesn't fix anything as evaluating `teal.code::chunks_new()` will warn anyway. I think, we need to:
- remove deprecation warnings from teal.code
- bump-up version to release teal.code
- reinstate deprecation warning and bump version of teal.code as a new main.

pawelru (approved):

PR #104, opened 2023-01-31
specify min deps

pawelru:
This PR specified minimal version requirement to pass the R CMD CHECK (incl. tests). If it is not specified then it means that it would work with **any** historical version which I have found not true.

This specification makes R CMD CHECK pass as well as make the _installation plan_ executable. E.g. `R6@2.0.0` works perfectly fine with the codebase but it won't be used as `testthat@2.3.0` requires `R6@2.2.0` so essentially this is the result of _truly installed_ minimal dependencies.


pawelru:
Hi @walkowif. I looked briefly into the workflow log and please find my comments:
- I can see that the install plan has need successfully resolved (that's the most problematic part) and it failed on the compilation error of not-so-old package 'testthat@2.3.0'. That's quite surprising. Can you check it please if it is installable on our CI image outside of this workflow?
- would it be possible (and does that make sense?) to split the script into the three steps executed one after another. The first one is resolving min deps. For debugging/troubleshooting we can call (uncomment) 'x$draw()' as well as add lock file as an artifact. Step 2 would be download and install (this is where we are failing now). This step would require object from step 1. Step 3 would be r cmd check of the current repo. What do you think about it?
- on a separate note: can we make this check non obligatory? At least for the very beginning until we make everything green. I am expecting lots of errors. 

I will have a deeper look tomorrow.

walkowif:
I've checked and running `docker run --rm ghcr.io/insightsengineering/rstudio_4.2.2_bioc_3.16:latest R -e  'install.packages("https://cran.r-project.org/src/contrib/Archive/testthat/testthat_2.3.0.tar.gz")'` returns the same compilation error as the one in the workflow.

We could split the script but I'm not sure what value this would add and I'm also not sure how to pass the object from step 1. to 2.
I moved and uncommented the lockfile creation and `draw` function, and ran the [workflow again](https://github.com/insightsengineering/teal.code/actions/runs/4126725196) - the lockfile is available as an artifact there.

The check can of course be made non-obligatory until it's tested properly.

PR #136, opened 2023-08-21
127 with(in) qenv

### LINKED ISSUE #127: Introduce `within.qenv`

```r
})
```

Update:

chlebowa:
Closes #127 

### CURRENT IMPLEMENTATION

_(NOTE: the new class is called `quenv` in the code to avoid conflicts with the existing `qenv`)._

`qenv` is a simple S3 class, an environment with some attributes: `code`, `errors`, `warnings`, `messages`.
`with.qenv` method is used to evaluate code in the `qenv`.

#### Syntax 

Just pass an expression to the `expr` argument in `with` call. 
Simple expressions should be enclosed in braces to store the code properly. Non-braced expressions are handled for convenience but it may be safer to use braces at all times.
```
q <- qenv()
with(q, {
  i <- iris
})
with(q, i <- iris)                 # alternative syntax
```

Compound expressions are supported but consecutive lines are evaluated separately, in turn.
```
with(q, {
  m <- mtcars
  ms <- subset(m, cyl == 4)
})
```

If a simple expression that is part of a compound expression raises an error, the following lines are not evaluated.
```
with(q, {
  w <- warpbreaks
  stop("that's far enough")        # stops evaluation
  w <- subset(w, wool == "B")     # not evaluated
})
```

Code can also be passed as strings, both literals and values, through the `text` argument.
```
xx <- "cc <- cars"
with(q, text = "cc <- cars")
with(q, text = xx)                 # equivalent to the above
```

Strings passed to the `expr` raise errors. Strings passed as parts of compound expressions are ignored.
```
with(q, "print(dim(m))")           # raises error
with(q, {"print(dim(m))"})         # not evaluated
with(q, text = "print(dim(m))")    # evaluated
with(q, print(dim(m)))             # equivalent to the above
```

The `text` argument can be used to evaluate code imported from a file.
```
with(q, text = readLines("code_file"))
```

#### Evaluation

Expression are evaluated in the `qenv` (per definition of the `with` generic it is called `data`) and the enclosing environment is the default `parent.frame()`. This includes the global environment in the scope, which makes the `i <- iris` calls possible.

In order to use expressions obtained programmatically, i.e. pass values that are not defined neither in `data`, nor in the parent frame, the expressions are first substituted in the current environment, i.e. the execution environment of the `with` call. Thus, we simply pass any extra values to the ellipsis:
```
q <- qenv()
with(
  data = q,
  expr = {
    i <- iris
    ii <- subset(i, Species == species)
  },
  species = "versicolor"
)
```
The same mechanism can utilize variables defined outside of the `with` call.
```
input_value <- "virginica"
q <- qenv()
with(
  data = q,
  expr = {
    i <- iris
    ii <- subset(i, Species == species)
  },
  species = input_value
)
```

#### Tracking `qenv` history

Each (single line) expression is stored in the `code` attribute, which is a list of expressions. After running the first code block, `attr(q, "code")` will be a list containing two expressions, both of them `i <- iris`.

The expression that is stored is the expression that is evaluated, after substitution of external values is complete. In the last two examples the subset expressions would be stored as `ii <- subset(i, Species == "versicolor")` and `ii <- subset(i, Species == "virginica")`, NOT as `ii <- subset(i, Species == species)` and `ii <- subset(i, Species == input_value)`, respectively.

#### Storing conditions

After an expression has been evaluated, messages from any conditions raised are stored in respective attributes, which are lists of character strings. 
If no condition is raised, an empty string is stored so that each condition list has the same length as `code` and condition messages can be easily matched to the code that raised them.

#### Utilities.

An elaborate `format` method prints a summary of contents, all expressions evaluated and all condition messages.
`get_code` and `get_condition` functions are provided for convenient retrieval of attributes.
Values are retrieved with the usual indexing operators, `$` and `[[`.

#### `within`

`within.qenv` works exactly like `with.qenv`, except it creates a deep copy of the `qenv` and returns the result.
```
q <- qenv()                        # create qenv
with(qenv, i <- iris)              # modify qenv
qq <- within(q, m <- mtcars)       # copy qenv, then modify and return the copy
```


__new_qenv.R
```diff
@@ -0,0 +1,316 @@
...
+    stop("specify either \"expr\" or \"text\": ", deparse1(parent_call), call. = FALSE)
+  }
+
+  if (!is.null(expr) && is.character(expr)) {
+    stop("character vector passed to \"expr\": ", deparse1(parent_call), call. = FALSE)
+  }
+
+  # Add braces to expressions. Necessary for proper storage of some expressions (e.g. rm(x)).
+  if (!is.null(expr) && !grepl("^\\{", deparse1(expr))) {
+    expr <- call("{", expr)
```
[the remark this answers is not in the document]
chlebowa: I **hate** `rlang` :wink: 

No, I haven't. Thanks for the suggestion.

__new_qenv.R
```diff
@@ -0,0 +1,316 @@
...
+  attr(ans, "messages") <- list()
+  class(ans) <- c("qenv", class(ans))
+  ans
+}
+
+
+#' @describeIn qenv act in `qenv` object
+#' @export
+with.qenv <- function(data, expr, text, ...) {
+  code <- .prepare_code(if (!missing(expr)) substitute(expr), if (!missing(text)) text)
```
averissimo, line 30: Would using `str2expression()` help simplify the processing of `text` expressions? There seem to be some issues with multi-line expression or compound expressions

Check these 2 examples (mocking reading the expression from file)

#### 1

```R
testthat::test_that("with.qenv Supports multiline text expression", {

  str_expr <- "1 +
  1"

  script_path <- withr::local_file(tempfile())
  writeLines(str_expr, script_path)

  q <- qenv()

  with(q, text = readLines(script_path)) |>
    testthat::expect_no_error()
})
```
#### 2

```R
testthat::test_that("with.qenv Supports multi-line expression", {

  str_expr <- "{
  1 +
  1
  2 * 3
}"

  script_path <- withr::local_file(tempfile())
  writeLines(str_expr, script_path)

  q <- qenv()

  with(q, text = readLines(script_path)) |>
    testthat::expect_no_error()
})
``

**Review by averissimo — COMMENTED** [review]
This looks really good!! :+1: 

I played around with it and found some (possible) issues with text expressions.

- [text param] Single expression spanning more than 1 line
- [text param] Compound expression  
- [text param] Ends with newline 

Regarding the first 2 I commented on the code, the last one is just a minor detail :sweat_smile: 

All 3 problems seem to go away when using `str2expression`, but I'm guessing there's a reason not to use that function

```R
testthat::test_that("with.qenv Supports text expression that snds with a newline", {

  str_expr <- "1 + 1
  2 * 3
"

  script_path <- withr::local_file(tempfile())
  writeLines(str_expr, script_path)

  q <- qenv()

  with(q, text = readLines(script_path)) |>
    testthat::expect_no_error()
})
```

chlebowa:
Thanks @averissimo! Since this is not a priority now I wasn't expecting a proper review yet so I'm happy to see some interest :)
Good work finding more edge cases, I'll have a look.
I did try going with `str2expression` but there was some reason against it which I now forget. I'll give it another try.

chlebowa:
Right, the reason I decided against `str2expression` is I wanted each line of compound expressions to be evaluated separately. That way if an error occurs, we can pinpoint the exact problem and display it to the user.
Theoretically this is also possible with `str2expression` as a compound expression can be disassembled but it is a pain to be able to handle simple things like `"1 + 1"`.

Perhaps I missed something. Where exactly did you use `str2expression`?

__new_qenv.R
```diff
@@ -0,0 +1,316 @@
...
+    text <- Filter(Negate(function(x) identical(x, "")), text)
+  }
+
+  code <-
+    if (is.null(text)) {
+      # Drop strings from compound expressions.
+      Filter(Negate(is.character), as.list(expr)[-1])
+    } else if (is.null(expr)) {
+      text
+    }
```
averissimo, line 271: A proposal for integrating the `str2expression`

```suggestion
  expr_text <- NULL

  if (!is.null(text)) {
    expr_text <- str2expression(text)

    if (length(expr_text) == 1 && isTRUE(expr_text[[1]][[1]] == "{")) {
      expr_text <- expr_text[[1]][seq(2, length(expr_text[[1]]))]
    }
  }

  code <-
    if (is.null(text)) {
      # Drop strings from compound expressions.
      Filter(Negate(is.character), as.list(expr)[-1])
    } else if (is.null(expr)) {
      Filter(Negate(is.character), as.list(expr_text))
    }
```

averissimo:
@chlebowa Sorry if I jumped the gun with the review.. this looks really good and I couldn't help myself to test it a bit :relaxed: 

I changed the `.prepare_code()` function by calling `str2expression()` on the `text` param.

ℹ️ See the comment on the code with suggestion on how to use it

But we can we actually run each expression separately, see toy example below

```R
text1 <- "1 + 2
3*3
a <- 2 +
3
a * 5"

text2 <- "{9 + 2
8*3
c <- 2 +
4
c * 10}"

cat_expr <- \(text) {
  expr <- str2expression(text)
  
  if (length(expr) == 0) return(invisible(NULL))
  
  if (length(expr) == 1 && isTRUE(expr[[1]][[1]] == "{")) {
    expr <- expr[[1]][seq(2, length(expr[[1]]))]
  }
  
  if (length(expr) > 1) {
    expr |> 
      purrr::walk(\(.x) cat("Expression:", deparse1(.x), "\n"))
  } else {
    # Otherwise treat it as single expression
    cat("Expression:", deparse1(expr[[1]]), "\n")
  }
  invisible(NULL)
}


cat_expr(text1)
cat_expr(text2)
cat_expr("1+1")
cat_expr("")
cat_expr("NULL")
cat_expr("NA")
```

chlebowa:
> @chlebowa Sorry if I jumped the gun with the review.. this looks really good and I couldn't help myself to test it a bit ☺️

By all means, be my guest :slightly_smiling_face: 

Ok, I see now that if the `else` block in the `code <- ` block is replaced with 
```
      text <- str2expression(text)
      if (length(text) == 1L && isTRUE(text[[1L]][[1L]] == "{")) {
        text <- text[[1L]][-1L]
      }
      as.character(text)
```
then the 
```
if (!is.null(text) && length(text) == 1L && grepl("^\\{", text)) {
    text <- strsplit(text, split = "\n")[[1]]
    text <- trimws(text, whitespace = "[ \t\r\n\\{\\}]")
    text <- Filter(Negate(function(x) identical(x, "")), text)
  }
``` 
above can be dropped.

Well done :+1: 

One question though: do we want to tolerate such cases (i.e. strings containing invalid expressions) or should we forbid them?

averissimo:
> One question though: do we want to tolerate such cases (i.e. strings containing invalid expressions) or should we forbid them?

What do you mean by invalid (text) expressions?

* bad syntax?
* valid multiline as the examples?
* others...

chlebowa:
I mean things like 
```
1 + 
1
```
that cause 
```
Error in str2lang("1 +") : <text>:2:0: unexpected end of input
1: 1 +
   ^
```

chlebowa:
You're right, I didn't consider loops and such.

chlebowa:
So, properly parsing all kinds of expressions (that came to mind) took some more effort bu t I think i got it right now. 

The idea is compound expressions are always disassembled into simple ones and those are evaluated sequentially, so that evaluation stops when an error is encountered.

On top of that, all simple expressions are stored (and printed) in the same neat form, regardless of how they were passed (braces, semicolons, line breaks, etc.).

R/with-qenv.R
```diff
@@ -0,0 +1,411 @@
...
+#' @export
+quenv <- function() {
+  ans <- new.env()
+  attr(ans, "code") <- list()
+  attr(ans, "errors") <- list()
+  attr(ans, "warnings") <- list()
+  attr(ans, "messages") <- list()
+  class(ans) <- c("quenv", class(ans))
+  ans
+}
```
gogonzo, line 93:
```suggestion
quenv <- function() {
  ans <- new.env()
  attr(ans, "code") <- list()
  attr(ans, "errors") <- list()
  attr(ans, "warnings") <- list()
  attr(ans, "messages") <- list()
  class(ans) <- c("quenv", class(ans))
  ans
}
```

Smart move with qenv being an environment. Not sure yet what is better:
1. qenv as S3 environment and protecting class by overwriting `$<-.qenv`, `[[<-.qenv` but potentially also other methods (which? `eval`?)
2. qenv as S4 and writing own `$`, `[[` methods to get objects.

I like `qenv` as environment, I think inheritance and reusing the class in other packages should be the key for a decision.

R/with-qenv.R
```diff
@@ -0,0 +1,411 @@
...
+
+
+#' @export
+#' @keywords internal
+`$<-.quenv` <- function(x, name, value) { # nolint
+  stop(
+    "Direct assignment is forbidden as it cannot be tracked. ",
+    "Use `with( <quenv>, { <name> <- <value> })` instead."
+  )
+}
```
gogonzo, line 231: instead of `$<-.quenv`, `[[<-.qenv` it is possible to just `lockEnvironment` and unlock only inside `within`

gogonzo (commented):
1. I like the idea with qenv being an S3 environment.
2. This `with.qenv` exposes us to some risks. 
3. Beware that qenv will be inherited in teal.data. Is it safe to use S3 in the way `class(x) <- c("tdata", "qenv", "environment")` and just register new methods for `tdata`?

gogonzo (commented):
To introduce `within` method we don't need to refactor whole package. We can add `within.qenv <- function(data, expr, text, ...)` function and decide some day about form of `qenv`.

R/with-qenv.R
```diff
@@ -0,0 +1,410 @@
...
+#'
+#' `quenv` creates `quenv` object.
+#' `with` acts in `quenv` object.
+#' `within` creates and modifies a (deep) copy of `quenv` object.
+#' `get_code` returns list of function calls or a data.frame with code and the conditions it raised.
+#' `get_conditions` returns list of condition messages (character strings).
+#'
+#' @param data,x (`quenv`)
+#' @param expr (`language`) simple or compound expression to evaluate in `data`
+#' @param text (`character`) character vector of expressions to evaluate in `data`
```
gogonzo, line 18:
```suggestion
#' @param code (`character`, `language`) object containing a valid R expression.
```

donyunardi:
Closing in favor of https://github.com/insightsengineering/teal.code/pull/149

PR #146, opened 2023-09-29
#133 code parser alternative

m7pr:
Fixes #133 

Alternative to https://github.com/insightsengineering/teal.code/pull/139

PR #147, opened 2023-10-02
127 introduce `within.qenv`

### LINKED ISSUE #127: Introduce `within.qenv`

chlebowa:
Closes #127 

Added an S3 `within` method for class `qenv`.
Usage: `within(data, expr, code, ...)`.
* argument `data` is `qenv`, this is handled by S3 dispatch
* argument `expr` is for literal code
* argument `code` is for variables holding code (call, list of calls, expression, list of expressions) or literal character strings

Code passed in any form is converted into list of (single) calls, which are evaluated sequentially.
The whole code is appended to the `@code` slot. Currently this is stored as `expression` but following #146 the slot will hold a single string, which will enable keeping comments.
Characters (vectors or strings) passed to `code` are stored as single strings, combined with `paste(..., collapse = "\n")`.
Language (calls or lists of calls, expressions or lists of expressions) passed to `code` are deparsed and combined likewise.


chlebowa:
Since we want to store code as is, it becomes an issue to perform substitution properly. Code coming in (whether it's through `expr` or `code`) is stored in the `@code` slot and only then converted into single calls which are substituted with values from `...` and evaluated.

chlebowa:
A solution has been proposed :+1:

PR #149, opened 2023-10-04
127 introduce `within.qenv`

### LINKED ISSUE #127: Introduce `within.qenv`

chlebowa:
Closes #127 

Added an `S3` `within` method for class `qenv` which is a wrapper for `eval_code`.

Usage: `within(data, expr, ...)`

`within.qenv` is created for convenience and accepts only inline expressions, not language or character variables, nor literal strings.

The `...` argument can be used to inject values into `expr` before evaluation.

#### EXAMPLES

Evaluate simple or compound expressions in `qenv`.
```
q <- new_qenv()
q <- within(q, i <- iris)
q <- within(q, {
  m <- mtcars
  f <- faithful
})
q
within(q, dim(f))
```

Inject value into `expr`.
```
qq <- within(q, ii <- subset(i, Species == species), species = "virginica")
within(qq, print(summary(ii)))
```

Inject value with variable.
```
input_value <- "versicolor"
qq <- within(q, ii <- subset(i, Species == species), species = input_value)
within(qq, print(summary(ii)))
```


chlebowa:
@gogonzo In the current form `within` cannot accept language objects or lists of language objects. If one really wants to squeeze an `expression` there, one can use `do.call`.
```
# single expression
expr <- expression(i <- iris, m <- mtcars)
do.call(within, list(q, expr))

# list of expressions
exprlist <- list(expression(i <- iris), expression(m <- mtcars))
do.call(within, list(q, do.call(c, exprlist)))
```

(Calls and call lists will not work, it has to be a single `expression`.)

The problem(?) is substitution is not available in these cases. The question is: do we care? Because I can make it work but then `within` itself will be open for directly passing `expression`s (possibly `call`s as well). I guess a use case would be something like what happens in tmc, where a module builds up a list of expressions and pushes them into a `qenv` all at once, but then the expressions are already properly substituted. On the other hand, the substitutions have to be done for each expression separately, whereas with this we could have them all done in one place, which would be neat, wouldn't it?

gogonzo (commented):
I think that warning is not needed for single symbol

R/qenv-within.R
```diff
@@ -0,0 +1,86 @@
...
+      call. = FALSE
+    )
+  }
+
+  # Add braces for consistency.
+  if (!identical(as.list(expr)[[1L]], as.symbol("{"))) {
+    expr <- call("{", expr)
+  }
+  # Drop strings from compound expressions.
+  calls <- Filter(Negate(is.character), as.list(expr)[-1])
```
gogonzo, line 80: Why to drop them? Let's keep "Garbage in, garbage out"

tests/testthat/test-qenv-within.R
```diff
@@ -0,0 +1,173 @@
...
+      1 + 1; 2 + 2
+    })
+  )
+  testthat::expect_no_error(
+    within(q, {
+      1 +
+        1
+    })
+  )
+})
```
gogonzo, line 35: This test in meaningless. All these are valid R code block expressions
```suggestion
```

tests/testthat/test-qenv-within.R
```diff
@@ -0,0 +1,173 @@
...
+    1 + 1;
+    2 + 2
+  })
+  all_code <- get_code(q)
+  all_code_pairs <- lapply(seq_len(4L), function(x) all_code[((x - 1L) * 2L) + 1:2])
+  testthat::expect_identical(
+    all_code,
+    rep(c("1 + 1", "2 + 2"), 4L)
+  )
+})
```
gogonzo, line 101: This test is too general and description doesn't explicitly says what is the code output. Feel free to write even more detailed than I've suggested.

```suggestion
testthat::test_that("within extracts content of the code block wrapped in curly brackets", {
  ....
})

testthat::test_that("within parses expr and ignores styling", {
  q <- within(q, {
    1 + 1
  })
  q <- within(q, {
    1 +
      1
  })
  q <- within(q, {
    1 + 1; 2 + 2
  })
    q <- within(q, {
    1 + 1; 
    2 + 2
  })
  q <- within(q, {
    1 + 1 
    2 + 2
  })
  testthat::expect_identical(...)
})
```

tests/testthat/test-qenv-within.R
```diff
@@ -0,0 +1,173 @@
...
+
+
+## injecting values ----
+testthat::test_that("external values can be injected into expressions through `...`", {
+  q <- new_qenv()
+
+  q <- within(q, {
```
gogonzo, line 108: This call has no meaning

tests/testthat/test-qenv-within.R
```diff
@@ -0,0 +1,173 @@
...
+      "iii <- subset(iris, Species == \"versicolor\")"
+    )
+  )
+
+  q <- new_qenv()
+  species <- "setosa"
+  qq <- within(q, {
+    i <- subset(iris, Species == species)
+  })
+  testthat::expect_s3_class(qq, "qenv.error")
```
gogonzo, line 137: this should be a separate test - "failure in the expr evaluation causes within to return qenv.error"

R/qenv-within.R
```diff
@@ -0,0 +1,86 @@
...
+  # Get parent call to use in error messages.
+  parent_call <- deparse1(match.call(definition = sys.function(1L), call = sys.call(1L)))
+
+  if (is.symbol(expr)) {
+    warning(
+      "In ", parent_call, " : Symbol passed to \"expr\" will be evaluated within qenv as is. ", "\n",
+      "This function is intended for literal expressions, for variables containing code use `eval_code` instead.",
+      call. = FALSE
+    )
+  }
```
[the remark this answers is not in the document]
chlebowa: Yeah, that's fair, let's assume the user knows what they are doing :+1:

gogonzo (commented):
Please try to write tests with single expectation and precise description.

R/qenv-within.R
```diff
@@ -0,0 +1,86 @@
...
+#' @param ... `name:value` pairs to inject values into `expr`
+#'
+#' @return
+#' Returns a `qenv` object with `expr` evaluated. If evaluation raises an error, a `qenv.error` is returned.
+#'
+#' @seealso [`eval_code`], [`base::within`]
+#'
+#' @export
+#'
+#' @rdname within
```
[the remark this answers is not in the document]
chlebowa: I don't believe it is possible to merge this method's docs into those of the `base` package. While this is not ideal, I was thinking we would tie `within` to `qenv`, see #150.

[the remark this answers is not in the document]
chlebowa: For the time being, I will add a note about `?within.qenv` in the vignette.

R/qenv-within.R
chlebowa: Fair enough.
It's a holdover from the time text was to be passed to a separate argument. I agree, let's keep it simple.

tests/testthat/test-qenv-within.R
gogonzo: Sorry for above, I understand that meaning of this line is different. I propose separate testthat "within doesn't look for an object in the parent frame when not provided in `...`" - or similar

tests/testthat/test-qenv-within.R
chlebowa: Modified to test for identity of substituted and unsubstituted expressions.

tests/testthat/test-qenv-within.R
chlebowa: Removed all but one braced expression.

tests/testthat/test-qenv-within.R
```diff
@@ -0,0 +1,117 @@
...
+  species = "virginica")
+
+  external_value <- "virginica"
+  q3 <- within(q, {
+    i <- subset(iris, Species == species)
+  },
+  species = external_value)
+
+  testthat::expect_identical(get_code(q1), get_code(q2))
+  testthat::expect_identical(get_code(q1), get_code(q3))
```
gogonzo, line 96: If the test reflected its description, then it would look as suggested. It would be shorter and more precise


```suggestion
  external_value <- "virginica"
  q <- within(
    new_qenv(), 
    i <- subset(iris, Species == species),
    species = external_value
   )

  testthat::expect_identical(get_code(q), 'i <- subset(iris, Species == "virginica")')
```

chlebowa: Isn't the `q2` case worth testing as well?

gogonzo: I don't think so. If we consider `q2` important, then we can also find another case. For example:

```r
foo <- function() "virginica"
within(iris$Species == species, species = foo())
```

gogonzo (approved):
We got it! 🔥 🔥 🔥

m7pr:
respect

PR #152, opened 2023-10-05
`format_expression` changes

chlebowa:
fruits of a fruitful discussion


tests/testthat/test-utils.R
```diff
@@ -60,5 +17,49 @@ testthat::test_that("format expression concatenates results of remove_enclosing_
...
     })
   )
-  expect_equal(format_expression(code_list), c("x <- 1", "y <- 1", "z <- 1"))
+  expect_equal(format_expression(code_list), c("x <- 1", "y <- 1\nz <- 1"))
+})
+
+
+# lang2calls ------------------------------------------------------------------------------------------------------
+
+testthat::test_that("format_expression returns expression/calls into characters without curly brackets", {
```
chlebowa, line 26:
```suggestion
testthat::test_that("format_expression turns expression/calls into character strings without curly brackets", {
```

tests/testthat/test-utils.R
```diff
@@ -60,5 +17,49 @@ testthat::test_that("format expression concatenates results of remove_enclosing_
...
+      return(x)
+    }
+  )
+
+  testthat::expect_identical(format_expression(expr1), "i <- iris\nm <- mtcars")
+  testthat::expect_identical(format_expression(expr2), c("i <- iris", "m <- mtcars"))
+  testthat::expect_identical(format_expression(expr3), c("expression(i <- iris)", "expression(m <- mtcars)"))
+  testthat::expect_identical(format_expression(cll1), c("i <- iris", "m <- mtcars")) # FAILS
+  testthat::expect_identical(format_expression(cll2), c("i <- iris", "m <- mtcars"))
+  testthat::expect_identical(
```
chlebowa, line 61: I thought they would all be `"i <- iris\nm <- mtcars"` :thinking: 
Will have a look tomorrow.

chlebowa:
`lang2calls` will enter an endless recursive loop if it encounters a string (and possibly other atomic vectors).
It needs either an assertion on `x` being `language`, which we don't want because we want to allow lists of calls, etc., or an exit condition like `if (is.atomic(x)) return(x)` at the very beginning. Or something else.

m7pr (approved):

m7pr:
@chlebowa I did provide `if (is.atomic(x)) return(x)` in `lang2calls` - thanks for the mentoring, help and the updates of tests. I will merge to the feature branch of code parser

PR #153, opened 2023-10-06
Introduce teal_data class

gogonzo:
Couple of fixes needed to use qenv in other packages:

Skärmavbild 2023-10-06 kl  10 43 48

https://cran.r-project.org/doc/manuals/R-exts.html#Namespaces-with-S4-classes-and-methods


pawelru:
Please add NEWS

vedhav (approved):
LGTM

chlebowa:
`methods::new` is no longer needed then, is it?

gogonzo:
> methods::new is no longer needed then, is it?
True. Not needed, but not suggested to do otherwise neither

Issue #150, opened 2023-10-04
[Question]: Should we clean the documentation of new_qenv?


m7pr:
Described in here https://github.com/insightsengineering/teal.code/pull/146#pullrequestreview-1654732556

chlebowa:
Speaking as the author of the original comment, I believe we should.
I would add `within` there as well, once #149 is merged.

m7pr:
@chlebowa should I rename to `qenv` instead of `new_qenv`. Is this issue stil valid?

chlebowa:
> @chlebowa should I rename to `qenv` instead of `new_qenv`. Is this issue stil valid?

No!

PR #156, opened 2023-10-24
145 create qenvs empty

### LINKED ISSUE #145: Don't allow to specify initial env and a code
`qenv` is a low level reproducibility tool. To make things simple as possible let's skip the case where we initialize `qenv` with specific `env` and `code`. Such "break" should be implemented only in the classes which inherit from `qenv`. Thanks to this `qenv` will be simpler and more robust.

chlebowa:
Closes #145 

I suggest we take this opportunity to also do #150


R/qenv-constructor.R
```diff
@@ -1,66 +1,21 @@
...
-#' @rdname new_qenv
-#' @export
-setMethod(
-  "new_qenv",
-  signature = c(code = "missing", env = "missing"),
-  function(env, code) {
-    new_qenv(env = env, code = code)
-  }
-)
+new_qenv <- function() {
```
m7pr, line 17: @gogonzo @chlebowa what about renaming to just `qenv()`?

gogonzo: +1

gogonzo: Feel free to do this in this PR - but please deprecate `new_qenv` with relevant message.

m7pr: Should we also deprecate parameters? Even though we are fine with a breaking change, we could warn that implementation does not use them at all.

```
new_qenv <- function(code, env) {
lifecycle::deprecate_warn("0.4.1.9006", "new_qenv(code)")
lifecycle::deprecate_warn("0.4.1.9006", "new_qenv(env)")

...

}
```

chlebowa: I didn't know that was even on the table :exploding_head:

m7pr:
Should we update NEWS + bump version, so that maybe there is a way to display info on deprecations https://github.com/insightsengineering/teal.code/pull/156/files#r1376050705?

R/qenv-constructor.R
gogonzo: `teal.code` is already on CRAN, so it is a must.

m7pr: deprecation of `new_qenv()` in favour of `qenv()` should be enough

chlebowa: Note that _renaming_ the constructor rather than just changing its arguments will have to be reflected in `teal.data` and possibly in other places.

m7pr: Yeah, there is plenty of places probably. would you like me to help with that?

R/qenv-constructor.R
```diff
@@ -2,20 +2,90 @@
...
+
+#' @export
+setMethod(
+  "new_qenv",
+  signature = c(env = "environment", code = "character"),
+  function(env, code) {
+    new_env <- rlang::env_clone(env, parent = parent.env(.GlobalEnv))
+    lockEnvironment(new_env, bindings = TRUE)
+    if (length(code) > 0) code <- paste(code, collapse = "\n")
+    id <- sample.int(.Machine$integer.max, size = length(code))
```
m7pr, line 67: @chlebowa should we put a deprecatnion statement in here?

```r
lifecycle::deprecate_warn("version", "new_qenv()", "qenv()")
```

R/qenv-constructor.R
chlebowa: Let's change the modules first, maybe then it only be `teal.data`.

R/qenv-constructor.R
chlebowa: I put one in the generic.

chlebowa:
Great catch @pawelru :clap:

pawelru:
1. There is one more `new_quenv` in docs: https://github.com/insightsengineering/teal.code/blob/7ba3f541b7f816174a8f9fe406a0b89867c9c35f/R/qenv-replace_code.R#L16
2. Please add NEWS entry. It's a big change.

tests/testthat/test-qenv_concat.R
```diff
@@ -1,21 +1,23 @@
 testthat::test_that("Concatenate two identical qenvs outputs", {
-  env <- new.env()
-  env$iris1 <- iris
-  q1 <- new_qenv(quote(iris1 <- iris), env = env)
+  q <- qenv()
+  q1 <- eval_code(q, quote(iris1 <- iris))
   q2 <- q1
 
   q <- concat(q1, q2)
```
pawelru, line 6: To avoid re-definition

```suggestion
  q0 <- qenv()
  q1 <- eval_code(q0, quote(iris1 <- iris))
  q2 <- q1

  q <- concat(q1, q2)
```

R/qenv-constructor.R
```diff
@@ -1,24 +1,52 @@
...
-#' @name new_qenv
+#' Create an empty `qenv` object.
 #'
-#' @param code (`character(1)` or `language`) code to evaluate. Accepts and stores comments also.
-#' @param env (`environment`) Environment being a result of the `code` evaluation.
+#' `qenv()` instantiates a `qenv` with an empty environment.
+#' Any changes must be made by evaluating code in it with `eval_code` or `within`, thereby ensuring reproducibility.
+#'
+#' `new_qenv()` can instantiate a `qenv` object with data in the environment and code registered.
+#' This method is deprecated.
```
gogonzo, line 9: Let's not give user a room to think for a second that this method could be used for something.
 
```suggestion
#' `new_qenv()` (deprecated) could instantiate a `qenv` object with data in the environment and code registered.
```

gogonzo (approved):
Clean

Issue #93, opened 2022-11-14
qenv is not reproducible - library calls


Polkas:
`qenv` is not reproducible as not taking into account libraries from its environment.

```
library(ranger)
qq <- teal.code::new_qenv() |> teal.code::eval_code("a <- ranger")
# code is not reproducible
teal.code::get_code(qq)
# But the code is evaluated properly
qq[["a"]]
```

It will be the best to not take blindly all loaded packages in the Session.
Possible we should not use the parent of Globalenv for default environment, and always demand library calls in the code.

gogonzo:
@Polkas yes I agree that packages from the session info is too much to show and many of the packages in the session are not related to execution within qenv. I think this is more `teal` issue than `teal.code`, as I don't really see the option within qenv to either check if @code has relevant library calls or if it's reproducible. So, I'd like to move it to teal if you agree.

`qenv` is reproducible or not depending on what user push. When one initializes `qenv` it would be a good idea to include library calls also and this is what we do in teal - but I agree, shole sessionInfo is too much. What to do then?

- in teal we can add library calls which are relevant to the code we are passing to tlg-modules
- in tlg-module we can add in the beginning library calls which are needed in the module


Any other ideas?

Polkas:
TL;DR I do not relate this issue in any way to teal or teal apps. qenv should always provide reproducible code or a proper warnings if not. The internal qenv libraries are NEUTRAL for a qenv search.

Because of the default new_env <- rlang::env_clone(env, parent = parent.env(.GlobalEnv)) we are on purpose offering all loaded in the session packages for the qenv. So we open the door for lack of reproducibility as later are not considered when get_code is used. Then it is a clear qenv problem for me. More than that we could not use libraries inside the qenv.

ranger is not loaded as the primary `parent.env(.GlobalEnv)` not known the ranger then.

```
# Restart the session
qq <- teal.code::new_qenv() |> teal.code::eval_code("library(ranger); a <- ranger")
# code is not reproducible
teal.code::get_code(qq)
# But the code is evaluated properly
qq[["a"]]
```

gogonzo:
@Polkas yes true. We might consider shifting libraries also, but I don't think it's the case as `eval_code` is evaluated within a package which already loads some libraries. If `library` call will be really executed, then I think we should do the shuffle.
Let me think more about this so I can prepare comprehensive answer

Polkas:
PROBLEM 1 lack of  library calls in get_code
[...]

PR #157, opened 2023-10-25
library calls possible within a call

gogonzo:
Possible to include `library` call in `eval_code`

https://github.com/insightsengineering/teal.code/issues/93

### DIFF
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -40,6 +40,9 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
       tryCatch(
         {
           eval(single_call, envir = object@env)
+          if (!identical(parent.env(object@env), parent.env(.GlobalEnv))) {
+            parent.env(object@env) <- parent.env(.GlobalEnv)
+          }
           NULL
         },
         error = function(e) {
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -41,6 +41,8 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
         {
           eval(single_call, envir = object@env)
           if (!identical(parent.env(object@env), parent.env(.GlobalEnv))) {
+            # needed to make sure that @env is always a sibling of .GlobalEnv
+            # could be changed when any new package is added to search path (through library or require call)
             parent.env(object@env) <- parent.env(.GlobalEnv)
           }
           NULL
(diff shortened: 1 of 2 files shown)


chlebowa:
What if someone detaches a package that si in the middle of the search path (for whatever reason)?

gogonzo:
> What if someone detaches a package that si in the middle of the search path (for whatever reason)?

as far as one uses `eval_code` to do so it's okey from reproducibility perspective

chlebowa:
But you can then lose a function from scope and the condition won't be triggered. I can't decide if that's going to be a problem :thinking:

gogonzo:
> But you can then lose a function from scope and the condition won't be triggered. I can't decide if that's going to be a problem 🤔

Removing namespace and then trying to obtain object from this namespace is a bug introduced by user. In the interactive session this would fail as well. It doesn't seem to be a `qenv` problem.

chlebowa:
Sounds good :+1:

chlebowa (approved):

chlebowa:
We haven't considered that using this feature forces us to call `teal.data` (in practice probably `teal`) before packages required for pre-processing.
Instead of 
```
  library(scda)
  library(scda.2022)
  library(dplyr)
  library(teal.modules.clinical)
  library(teal.modules.general)
  library(teal.osprey)
  library(nestcolor)
  # optional libraries
  library(sparkline)
```
one would do 
```
  library(teal)
  library(scda)
  library(scda.2022)
  library(dplyr)
  library(teal.modules.clinical)
  library(teal.modules.general)
  library(teal.osprey)
  library(nestcolor)
  # optional libraries
  library(sparkline)
```

This leaves us with a different search path:
```
                             old                           new
1                     .GlobalEnv                    .GlobalEnv
2              package:sparkline             package:sparkline
3              package:nestcolor             package:nestcolor
4            package:teal.osprey           package:teal.osprey
5                 package:osprey                package:osprey
6   package:teal.modules.general  package:teal.modules.general
7              package:shinyTree             package:shinyTree
8               package:ggmosaic              package:ggmosaic
9                package:ggplot2               package:ggplot2
10 package:teal.modules.clinical package:teal.modules.clinical
11                  package:tern                  package:tern
12               package:rtables               package:rtables
13            package:formatters            package:formatters
14                  package:teal                 package:dplyr
15        package:teal.transform             package:scda.2022
16              package:magrittr                  package:scda
17            package:teal.slice                  package:teal
18             package:teal.data        package:teal.transform
19                 package:shiny              package:magrittr
20                 package:dplyr            package:teal.slice
21             package:scda.2022             package:teal.data
22                  package:scda                 package:shiny
23                 tools:rstudio                 tools:rstudio
24                 package:stats                 package:stats
25              package:graphics              package:graphics
26             package:grDevices             package:grDevices
27                 package:utils                 package:utils
28              package:datasets              package:datasets
29               package:methods               package:methods
30                     Autoloads                     Autoloads
31                  package:base                  package:base
```

I don't think this should be a problem but it is something we should be aware of.

PR #160, opened 2023-11-07
add code masking

chlebowa:
Added private function `replace_code` to facilitate manipulating contents of `@code`.


chlebowa:
Frankly, the more I look at the example, the more uncomfortable I feel exporting this function.

chlebowa:
The function name is up for debate. I even considered `set_code` but that would imply that the whole slot can be modified.

R/qenv-replace_code.R
```diff
@@ -0,0 +1,53 @@
...
+#'
+#' @return A modified `qenv`.
+#'
+#' @seealso [`qenv-class`], [`eval_code`]
+#'
+#' @examples
+#' q <- new_qenv()
+#' q <- eval_code(q, expression(i <- iris, ii <- subset(i, Species == "virginica")))
+#' get_code(q)
+#' q <- replace_code(q, expression(i <- iris, ii <- subset(i, Species != "virginica")))
```
gogonzo, line 19: Replacing code with the same code? What do you think about this?
```suggestion
#' q <- eval_code(q, expression(print("secretpassword"))
#' get_code(q)
#' q <- replace_code(q, expression(print(askpass::askpass()))
```

chlebowa: It wasn't the same though.

tests/testthat/test-qenv-replace_code.R
```diff
@@ -0,0 +1,63 @@
...
+})
+
+testthat::test_that("non-last elements of @code are unaffected", {
+  qq <- within(qq, m <- mtcars)
+  replacement <- "i <- mtcars"
+  qr <- replace_code(qq, replacement)
+  before <- get_code(qq)[-length(get_code(qq))]
+  after <- get_code(qr)[-length(get_code(qr))]
+  testthat::expect_identical(before, after)
+})
```
gogonzo, line 57: Replacement of the last element of the @code is exactly what `code_replace` does and other tests checking derivative information (length, last different, non-last identical)

```suggestion
testthat::test_that("code_replace replaces last element of the code", {
  qq <- within(qq, m <- mtcars)
  replacement <- "i <- mtcars"
  qr <- replace_code(qq, replacement)
  previous <- get_code(qq)
  current <- get_code(qr)
  testthat::expect_identical(current, c(head(previous, -1), replacement)
})
```

R/qenv-replace_code.R
```diff
@@ -0,0 +1,53 @@
+#' Replace code
+#'
+#' Replace last element of code stored in `qenv`.
+#'
+#' Replace the last element of the character vector stored in `@code` with provided `code`.
+#' This can be used to mask sensitive information, e.g. credentials.
```
pawelru, line 6: When someone from the outside will be looking at this then this functionality might be super suspicious. Can you please provide few words about the context - i.e. hiding tokens, passwords etc? I know that you have this already in vignette but that someone will be probably looking only here.

chlebowa: That's what worries me, though. Our intentions are rather pure and we can dress it up in pretty examples but the fact is you can put whatever you want in there.

Also, doesn't "credentials" cover what you mean?

pawelru: > Also, doesn't "credentials" cover what you mean?

Yes

_pkgdown.yml
```diff
@@ -27,6 +27,7 @@ reference:
       - concat
       - dev_suppress
       - eval_code
+      - replace_code
```
gogonzo, line 30: Internal docs shouldn't be listed in pkgdown.yml, please remove

gogonzo (approved):
Good to go. Please remove replace_code from pkgdown

_pkgdown.yml
chlebowa: The one I always forget :roll_eyes:

Issue #173, opened 2023-11-27
get_code to return `character(1)` instead of `character(n)`


gogonzo:
- `get_code` should return single string with all @code concatenated to a single string.

Please make relevant changes:
- We don't need to use `paste(get_code(...), collapse = "\n")` everywhere anymore. Please locate all `paste(collapse = "\n")` in `teal.modules.xyz` and in `teal` where code is concatenated to a single string.

chlebowa:
The internal function `format_expression` can also be removed. It is used in very few places and that number will soon drop further.

vedhav:
Maybe I'm missing something. Is there some reason that this issue is bucked into the wider `refactor` scope? I think it's fine to merge this directly into `main` without going through `refactor`.

PR #176, opened 2023-11-27
`get_code` to return concatenated code string and removed `format_expression`

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal.code/issues/173

In this pull request, I've made updates to both the `get_code` and `replace_code` methods. Now, when the `deparse `arg is set to TRUE, these methods will return a concatenated string. Specifically, the `get_code` method will return character(0) when used with `new_qenv(),` provided no code has been set. As for `replace_code,` it will first split the input string using "\n", then replace the last string in this sequence. Finally, it concatenates these strings again before replacing the code.

Removed `format_expression` function.

Example
```

q <- new_qenv()
get_code(q) # return character(0)
replace_code(q, "i <- iris")  # @code return character(0)

qq <- within(q, i <- iris)
qq <- within(qq, m <- mtcars)
get_code(qq) # return "i <- iris\nm <- mtcars"

replacement <- "i <- mtcars"
qr <- replace_code(qq, replacement)
get_code(qr) #return "i <- iris\ni <- mtcars"
```

Also review following modules for PR for changes.

- [x] teal https://github.com/insightsengineering/teal/pull/976
- [x] teal.data https://github.com/insightsengineering/teal.data/pull/206
- [x] teal.modules.general https://github.com/insightsengineering/teal.modules.general/pull/615
- [x] teal.modules.clinical https://github.com/insightsengineering/teal.modules.clinical/pull/898
- [x] teal.goshawk https://github.com/insightsengineering/teal.goshawk/pull/250
- [x] teal.osprey https://github.com/insightsengineering/teal.osprey/pull/244

### DIFF
--- a/R/qenv-get_code.R
+++ b/R/qenv-get_code.R
@@ -33,7 +33,7 @@ setMethod("get_code", signature = "qenv", function(object, deparse = TRUE) {
       paste(object@code, collapse = "\n")
     }
   } else {
-    parse(text = object@code, keep.source = TRUE)
+    parse(text = paste(c("{", object@code, "}"), collapse = "\n"), keep.source = TRUE)
   }
 })
 
--- a/tests/testthat/test-qenv_get_code.R
+++ b/tests/testthat/test-qenv_get_code.R
@@ -21,7 +21,7 @@ testthat::test_that("get_code returns expression of qenv object if deparse = FAL
   q <- eval_code(q, quote(y <- x))
   testthat::expect_equivalent(
     toString(get_code(q, deparse = FALSE)),
-    toString(parse(text = q@code, keep.source = TRUE))
+    toString(parse(text = paste(c("{", q@code, "}"), collapse = "\n"), keep.source = TRUE))
   )
 })
 


NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * Exported the `qenv` class from the package.
 * The `@code` field in the `qenv` class now holds `character`, not `expression`.
+* The `get_code` method  returns a single concatenated string of the code.
+* Removed internal `format_expression` function.
```
vedhav, line 8: I don't think removing `format_expression` should be added to `NEWS.md`. It's not a user-facing function and it will just clutter the actual news.
```suggestion
```

R/qenv-replace_code.R
```diff
@@ -32,19 +32,24 @@ setGeneric("replace_code", function(object, code) standardGeneric("replace_code"
...
 #' @keywords internal
 setMethod("replace_code", signature = c("qenv", "language"), function(object, code) {
-  replace_code(object, code = format_expression(code))
+  replace_code(object, code = paste(lang2calls(code), collapse = "\n"))
 })
 
 #' @keywords internal
 setMethod("replace_code", signature = c("qenv", "expression"), function(object, code) {
-  replace_code(object, code = format_expression(code))
+  replace_code(object, code = paste(lang2calls(code), collapse = "\n"))
```
vedhav, line 52: I'm having second thoughts about removing `format_expression`. It's used in multiple places. I think it can stay.

vedhav (changes requested):
I just have one comment regarding the changes for this issue. Other than that, I think everything looks fine to me. It's good that this issue is not a breaking change, so I believe updating just the `NEWS.md` is sufficient.

Could you please change the target branch to `main` instead of `refactor` in this and all the related child PRs? I think this issue is independent of the `refactor` changes, and we can merge it before or after the `refactor` without any issues.

R/qenv-replace_code.R
kartikeyakirar: I had a similar thought previously, but following a conversation with @chlebowa it was clarified that the function simply uses paste with a designated collapse parameter after processing with `lang2calls`. The `format_expression` function originally had a more complex design but it was simplified into a simple one-liner in a past refactoring phase. Currently, it's essentially redundant.

R/qenv-get_code.R
```diff
@@ -27,7 +27,11 @@ setGeneric("get_code", function(object, deparse = TRUE, ...) {
 setMethod("get_code", signature = "qenv", function(object, deparse = TRUE) {
   checkmate::assert_flag(deparse)
   if (deparse) {
-    object@code
+    if (length(object@code) == 0) {
+      object@code
+    } else {
+      paste(object@code, collapse = "\n")
+    }
```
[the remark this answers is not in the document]
kartikeyakirar: I think it's a good idea, but after looking into it, I found that the `qenv` class has a `setValidity` function that enforces the lengths of the `@code`, `@id`, `@warnings`, and `@messages` slots to be the same. If we concatenate all the executed code into a single string and store it in the `@code` slot, it will affect the length comparison checks. 

To maintain the integrity of the `qenv` object, it might be better to keep it as a list or vector instead of concatenating it into a single string.

R/qenv-replace_code.R
m7pr: format_expression sounds like a wrapper for paste, where you overwrite default collapse param :)

m7pr:
@vedhav 

> Could you please change the target branch to main instead of refactor in this and all the related child PRs? I think this issue is independent of the refactor changes, and we can merge it before or after the refactor without any issues.

We also have `get_code.teal_data` that exists only on `@refactor` in `teal.data` so I would keep this aligned with `@refactor` branch.

m7pr:
@kartikeyakirar why it only shortens the length for `deparse = TRUE`? For `deparse=FALSE` it returns an expression, but it could return an expression of length 1 as well as it returns pasted code into length 1 right?

R/qenv-get_code.R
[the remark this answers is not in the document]
m7pr: because of this `setValidity` I initially though `@code` equal to the number of applied `evaluations` was something desired

kartikeyakirar:
> why it only shortens the length for deparse = TRUE? For deparse=FALSE it returns an expression, but it could return an expression of length 1 as well as it returns pasted code into length 1 right?

For deparse=`FALSE` it returns an expression not necessarily of length 1, The choice not to collapse the expressions into a single expression (even if possible) due to maintain integrity of the code. Collapsing everything into a single expression could alter the behavior or interpretation of the code when it's later evaluated.

Here expression is of length 3
```
q1 <- new_qenv(env = list2env(list(a = 1)), code = quote(a <- 1))
q2 <- eval_code(q1, code = quote(b <- a))
q3 <- eval_code(q2, code = quote(d <- 2))
length(get_code(q3)) # 1
length(get_code(q3, deparse = FALSE)) # 3
```
even if you collpase it before converting it into expression result is same
```
expression <- parse(text = get_code(q3), keep.source = TRUE)
length(expression) # 3


```

m7pr:
@kartikeyakirar yeah, because you need to paste `code` with `{` and `}` to make it one call, hence an expression of length 1

```r
> code <- c('a<-1', 'b<-2', 'c<-2')
> length(parse(text = code))
[1] 3
> length(parse(text = paste(code, collapse = '\n')))
[1] 3
> length(parse(text = paste(c('{', code, '}'), collapse = '\n')))
[1] 1
```

kartikeyakirar:
> because you need to paste code with { and } to make it one call, 

yes. It won't alter the interpretation of the code when it's later evaluated. Why didn't I think of this? I will update test and make change in teal as well. thanks

R/utils.R
```diff
@@ -24,12 +24,6 @@ dev_suppress <- function(x) {
   force(x)
 }
 
-format_expression <- function(code) {
-  code <- lang2calls(code)
-  paste(code, collapse = "\n")
-}
-
```
averissimo, line 31: Did you consider having `lang2calls` return the character string itself?

It seems that it is only used this way in the package _(other than in tests)_

ps. and rename it to `lang2str` or `lang2character`?

R/qenv-replace_code.R
averissimo: Did you guys consider having `lang2calls` return the `paste(...)`?

It seems that it is only used this way in the package _(other than in tests)_

ps. and rename it to `lang2str` or `lang2character`?

kartikeyakirar: there is recursion in `lang2calls`  and  implementing the paste function with collapse directly inside `lang2calls` can lead to unintended results due to concatenation of results from each recursive call.

vedhav (approved):
I don't have any other comments, Good job in reflecting the change in all repos. LGTM

PR #179, opened 2023-11-29
Remove residual code and their dependencies

### LINKED ISSUE #158: remove `teal.code`'s dependency
`teal.code` `Imports` `shiny`. The only use seems to be 
```
include_css_files <- function(pattern = "*") {
  css_files <- list.files(
    system.file("css", package = "teal.code", mustWork = TRUE),
    pattern = pattern, full.names = TRUE
  )
  if (length(css_files) == 0) {
    return(NULL)
  }

  shiny::singleton(lapply(css_files, shiny::includeCSS))
}
```

I say we drop `shiny` from dependencies. Please review other dependencies if we can drop any.

vedhav:
Closes #158


kartikeyakirar (approved):
LGTM 👍

PR #183, opened 2023-12-11
165 trim trace for qenv.error@main

### LINKED ISSUE #165: code storage in `qenv.error`
The trace displayed by `get_code` called on `qenv.error` doesn't look right.

```
> qe <- qenv() |> within({IRIS = iris}) |> within(stop("oh no"))
> get_code(qe)
Error: oh no 
 when evaluating qenv code:
stop("oh no")

trace: 
 c("IRIS = iris", "stop(\"oh no\")")
```

I think when we changed `qenv@code` from `language` to `character` we forgot to adjust `qenv.error`.

m7pr:
Close #165


R/qenv-get_code.R
```diff
@@ -50,7 +50,7 @@ setMethod("get_code", signature = "qenv.error", function(object, ...) {
       sprintf(
         "%s\n\ntrace: \n %s\n",
         conditionMessage(object),
-        paste(lang2calls(object$trace), collapse = "\n ")
```
pawelru, line 53: if this is correct (which I am not quite sure yet) - please track down other usages of that function with the working hypothesis that this is the only one

m7pr: There is a lot, as we used this function to divide long expressions/languages into calls. Currently `object@trace` is equal to  `object@code` up to the last possible element of if that did throw the error. lang2calls is the successor of `format_expression` that we used to clean up expressions from `{`. I think we did to much quick substitutions of `format_expression` with `lans2calls` without further investigation and now we need to get rid of it in places where this does not belong.

chlebowa: > if this is correct (which I am not quite sure yet) - please track down other usages of that function with the working hypothesis that this is the only one

@pawelru I don't quite understand what you mean. Is there a problem with `lang2calls` itself?

pawelru: I meant to check if this is the only occurrence of this function. If it happens to be true then we can remove importing / defining it. But if there are other occurrences then no action needed.

chlebowa: I believe it's quite necessary when _storing_ code, which is why it can be omitted when retrieving it.

pawelru: then please ignore my comment

chlebowa (approved):
```
> devtools::load_all()
ℹ Loading teal.code
> qe <- qenv() |> within(i <- iris) |> within(stop("oh no"))
> qe
<qenv.error: oh no 
 when evaluating qenv code:
stop("oh no")>
> get_code(qe)
Error: oh no 
 when evaluating qenv code:
stop("oh no")

trace: 
 i <- iris
 stop("oh no")
```

This is what I intended :+1: 


As long as you're here, would you mind modifying this line in `get_code` like it was changed in `teal_data::get_code`?
```
# from
parse(text = paste(c("{", object@code, "}"), collapse = "\n"), keep.source = TRUE)
# to
parse(text = object@code, keep.source = TRUE)
```

I disabled auto-merge to give you chance.

m7pr:
Hey @chlebowa why would we want to change this?

image

The line 

```r
parse(text = paste(c("{", object@code, "}"), collapse = "\n"), keep.source = TRUE)
```

was introduced in here https://github.com/insightsengineering/teal.code/pull/176 so that resulted expression (for `deparse = FALSE`) has a length 1, so it's consistent with the length of the result for `deparse  = TRUE` https://github.com/insightsengineering/teal.code/pull/176#issuecomment-1829333986

chlebowa:
before:
```
> devtools::load_all()
ℹ Loading teal.code
> qe <- qenv() |> within(i <- iris) |> within(m <- mtcars)
> get_code(qe, deparse = TRUE)
[1] "i <- iris\nm <- mtcars"
>
```

after:
```
> devtools::load_all()
ℹ Loading teal.code
> qe <- qenv() |> within(i <- iris) |> within(m <- mtcars)
> get_code(qe, deparse = TRUE)
[1] "i <- iris\nm <- mtcars"
> 
```

m7pr:
If we are trying to be consistent with what's in https://github.com/insightsengineering/teal.data/blob/main/R/teal_data-get_code.R#L60 then I would actually modify what's in teal.data::get_code.teal_data so return length 1 for each deparse = FALSE

m7pr:
`get_code` in `teal_data` is inconsistent in terms of the length of the output if you want to have a deparsed or non-deparsed output

```r
library(teal.data)
tdata1 <- teal_data()
tdata1 <- within(tdata1, {
  a <- 1
  b <- a^5
  c <- list(x = 2)
})

length(get_code(tdata1, deparse = FALSE))
# > [1] 3
length(get_code(tdata1, deparse = TRUE))
# > [1] 1
```

chlebowa:
I stand corrected, I ran my tests wrong.

Merge away.

m7pr:
No worries, I made an issue in `teal.code` to follow-up on this one https://github.com/insightsengineering/teal.data/issues/224

PR #185, opened 2023-12-14
collate seealso fields

chlebowa:
Moved all seealso tags that are used in the qenv page to the qenv.R file so that the field is neatly fit into a single line.


R/qenv-constructor.R
```diff
@@ -47,7 +47,7 @@ qenv <- function() {
 #' @aliases new_qenv,environment,missing-method
 #' @aliases new_qenv,missing,missing-method
 #'
-#' @seealso [`get_var()`], [`get_warnings()`], [`join()`], [`concat()`]
+#' @seealso [`base::within()`], [`get_var()`], [`get_warnings()`], [`join()`], [`concat()`]
```
averissimo, line 50: The backticks are not necessary for function calls. They generate the same exact documentation

```suggestion
#' @seealso [base::within()], [get_var()], [get_warnings()], [join()], [concat()]
```

averissimo (approved):
Looks good. 1 comment that is optional.

R/qenv-constructor.R
chlebowa: Neat but I will keep the backticks in case I forget the parentheses at some point.

PR #186, opened 2023-12-14
94 `get_var` `inherits = FALSE`

### LINKED ISSUE #94: rethink get with inherits = TRUE
By mistake in https://github.com/insightsengineering/teal.code/issues/93 I subset q instead of a variable and find out another interesting thing.
We could use [[ in the recursive manner now:

```
qq <- teal.code::new_qenv() 
ls(qq@env) # is empty
# we get functions from lower environments
qq[["plot"]]
qq[["list.files"]]
```

It comes from the https://github.com/insightsengineering/teal.code/blob/f9f8cf9d0a4882f54d3a40faaa295979a654ca95/R/qenv-get_var.R#L23 where we use get instead of mget (or the inherits = TRUE). It is not a problem for me but an interesting functionality.

@gogonzo

chlebowa:
Closes #94 
Added `inherits = FALSE` to `get` call in `get_var`.
Added unit tests.

### DIFF
--- a/tests/testthat/test-qenv_get_var.R
+++ b/tests/testthat/test-qenv_get_var.R
@@ -24,3 +24,15 @@ testthat::test_that("get_var and `[[` return NULL if object not in qenv environm
   testthat::expect_null(q[["w"]])
   testthat::expect_message(q[["w"]], "object 'w' not found")
 })
+
+testthat::test_that("get_var and `[[` only returns objects from qenv, not parent environment(s)", {
+  q <- qenv()
+
+  testthat::expect_null(get_var(q, "iris"))
+  testthat::expect_null(q[["iris"]])
+
+  qq <- within(q, iris <- head(iris))
+
+  testthat::expect_s3_class(get_var(qq, "iris"), "data.frame")
+  testthat::expect_identical(dim(get_var(qq, "iris")), c(6L, 5L))
+})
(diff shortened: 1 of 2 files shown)


chlebowa:
Questions: 
1. Should `get_var` return `NULL` or raise error when variable does not exist? Currently prints message and returns `NULL`.
2. Should we remove `get_var` altogether as per [this comment](https://github.com/insightsengineering/teal.code/issues/94#issuecomment-1313766307)? May require a separate issue.

tests/testthat/test-qenv_get_var.R
```diff
@@ -24,3 +24,15 @@ testthat::test_that("get_var and `[[` return NULL if object not in qenv environm
...
+  q <- qenv()
+
+  testthat::expect_null(get_var(q, "iris"))
+  testthat::expect_null(q[["iris"]])
+
+  qq <- within(q, iris <- head(iris))
+
+  testthat::expect_s3_class(get_var(qq, "iris"), "data.frame")
+  testthat::expect_identical(dim(get_var(qq, "iris")), c(6L, 5L))
+})
```
gogonzo, line 38: I think these are two separate tests:
1. not accessing variables from parent envs
2. prioritizing this env over parents

However, I find second one irrelevant. As it is obvious that objects in this environments are accessible by `get_var` - expecialy that there is no other environments in the scope.

```suggestion
testthat::test_that("get_var and `[[` does not return objects from parent environments(s)", {
  q <- qenv()

  testthat::expect_null(get_var(q, "iris"))
  testthat::expect_null(q[["iris"]])
})

testthat::test_that("????", {
  q <- within(qenv(), iris <- head(iris))
  testthat::expect_identical(get_var(q, "iris"), head(iris))
})
```

chlebowa: Fair enough, removed.

gogonzo (approved):

PR #187, opened 2023-12-15
175 ls(qenv)

### LINKED ISSUE #175: [Feature Request]: Method to list names of the objects in `@env`.
### Feature description

We need a method to get names from @env to not force users to get directly from the slot `ls(<qenv>@env)`. Ideally we would like to use `ls(<qenv>)` or `objects(<qenv>)` but these functions are not registered generics in base.
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

m7pr:
Close #175


R/qenv-get_env.R
```diff
@@ -0,0 +1,32 @@
+#' Extract Environment Included in `qenv`
+#'
+#' The extraction of environment included in `qenv@env` allows to list object names included in `qenv@env` slot.
+#'
+#' @param object (`qenv`)
+#'
+#' @return A `character` vector with names of objects.
```
pawelru, line 7: nope

NAMESPACE
```diff
@@ -7,6 +7,7 @@ export(concat)
 export(dev_suppress)
 export(eval_code)
 export(get_code)
+export(get_env)
```
pawelru, line 10: Please add news entry. It's a prominent change

R/qenv-get_env.R
```diff
@@ -0,0 +1,32 @@
+#' Extract Environment Included in `qenv`
+#'
+#' The extraction of environment included in `qenv@env` allows to list object names included in `qenv@env` slot.
```
pawelru, line 3: - "Access"? An "extract" feels too complicated to me.
- (...) allows to **e.g.** list object names

R/qenv-get_env.R
m7pr: I wrote a documentation for ls() and then decided to leave it as get_env : D 4e8aa987d279ce694b657eb5ee212191f7f440fb thanks - fixed

NAMESPACE
m7pr: cc8bcc227bc2a6a804974aec90e4bebcf8588c45

R/qenv-get_env.R
m7pr: Access is fine :)
5d32b50c44b845d84975a1a808f182a3f9eab67a

NEWS.md
```diff
@@ -10,6 +10,7 @@
 * The `@code` field in the `qenv` class now holds `character`, not `expression`.
 * The `get_code` method  returns a single concatenated string of the code.
 * Added `within` support for `qenv.error` class.
+* Added `get_env,teal_data` method that allows to extract environment stored in `teal_data@env` slot.
```
pawelru, line 13: the term "teal_data" is unknown in the context of `teal.code` (it's in teal.data). Looking at the above entries we can skip class name to keep it simpler.

m7pr: a91bcc25bd936ff33f332403fab0cd7beb160a39

pawelru (approved):
This is good from my perspective and I like the approach for exposing slot for further use.
Please just raise awareness of this within the team as there is a conflicting PR.

pawelru:
I like this that much that I would even go with this to the README and vignettes. Please just make sure to align what's printed to avoid e.g. the following
```
qenv_2
#> <environment: 0x00000135b544cfe8> [L]
#> Parent: <environment: package:teal.code>
#> Bindings:
#> • x: <dbl> [L]
#> • y: <dbl> [L]
#> • z: <dbl> [L]

get_env(qenv_2)
<some other hash printed>
```

m7pr:
>This is good from my perspective and I like the approach for exposing slot for further use.
> Please just raise awareness of this within the team as there is a conflicting PR.

Just marked this solution in here https://github.com/insightsengineering/teal/pull/1004#discussion_r1427716753

m7pr:
@pawelru 

> Please just make sure to align what's printed to avoid e.g. the following

What exactly we should avoid? Printing `qenv` at all or what do you mean?

chlebowa:
I think @pawelru wanted to make sure the same env hash is printed but that is a given because `get_env` just takes the env from the slot and returns it as is.

m7pr:
Yeah, this crossed my mind but just wanted to double check

m7pr:
I think this might be an overkill. So `qenv` has a `show` method (that works as a print for S4 objects) that is implemented as

```r
setMethod("show", "qenv", function(object) {
  rlang::env_print(object@env)
})
```

So for `qenv` we basically have a nice way of printing what's in `qenv@env`. And now we have a `get_env` for which we would like to get another nice print? So `get_env` should return an object of `some class` for which we would use `rlang::env_print(` for it's print or show methods. I think that is too much.

chlebowa:
I believe `get_env` should return an environment (which it does) and there is a built-in method to print that. I wouldn't change anything else here.

m7pr:
Amazing, I did extend README and vignettes. Examples in README did not work as `qenv()` does not allow to specify any arguments anymore so it's great we had a chance to test :)

PR #191, opened 2024-01-10
Removing old rd syntax

kartikeyakirar:
this is part of https://github.com/insightsengineering/teal.code/pull/189

I noticed some old syntax that was not previously checked, so I am creating a separate pull request for it.

### DIFF
--- a/R/qenv-join.R
+++ b/R/qenv-join.R
@@ -12,14 +12,9 @@
 #'    Example:
 #'
 #'    ```r
-#'    x <- qenv(
-#'      code = c(mtcars1 = "mtcars1 <- mtcars"),
-#'      env = list2env(list(mtcars1 = mtcars))
-#'    )
-#'    y <- qenv(
-#'      code = c(mtcars1 = "mtcars1 <- mtcars['wt']"),
-#'      env = list2env(list(mtcars1 = mtcars['wt']))
-#'    )
+#'    x <- eval_code(qenv(), expression(mtcars1 <- mtcars))
+#'    y <- eval_code(qenv(), expression(mtcars1 <- mtcars['wt']))
+#'
 #'    z <- join(x, y)
 #'    # Error message will occur
 #'    ```
@@ -32,7 +27,7 @@
 #'
 #'    Example:
 #'    ```r
-#'    common_q <- qenv(code = "v <- 1", env = list2env(list(v = 1)))
+#'    common_q <- eval_code(qenv(), expression(v <- 1))
 #'    x <- eval_code(
 #'      common_q,
 #'      "x <- v"
--- a/R/qenv-join.R
+++ b/R/qenv-join.R
@@ -7,8 +7,10 @@
 #' See below for an example.
 #'
 #' There are some situations where `join()` cannot be properly performed, such as these three scenarios:
-#' 1. Both `qenv` objects contain an object of the same name but are not identical.\cr\cr
+#' 1. Both `qenv` objects contain an object of the same name but are not identical.
+#'
 #'    Example:
+#'
 #'    ```r
 #'    x <- qenv(
 #'      code = c(mtcars1 = "mtcars1 <- mtcars"),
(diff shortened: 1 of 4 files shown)


R/qenv-join.R
```diff
@@ -7,102 +7,102 @@
...
-#'   )
-#'   z <- eval_code(
-#'     y,
-#'     "z <- v"
-#'   )
-#'   q <- join(x, y)
-#'   join_q <- join(q, z)
-#'   # Error message will occur
+#' 1. Both `qenv` objects contain an object of the same name but are not identical.\cr\cr
+#'    Example:
```
averissimo, line 11: How about ditching the `\cr` as well?

```suggestion
#' 1. Both `qenv` objects contain an object of the same name but are not identical.
#'
#'    Example:
#'
```

R/qenv-join.R
```diff
@@ -7,102 +7,102 @@
...
-#'      for (i in 1:3) {
-#'        y <- c(y, i)
-#'      }"
-#'   )
-#'   q <- join(x,y)
-#'   # Error message will occur
+#' 2. `join()` will look for identical `@id` values in both `qenv` objects.
+#'    The index position of these `@id`s must be the same to determine the evaluation order.
+#'    Otherwise, `join()` will throw an error message.\cr\cr
+#'    Example:
```
averissimo, line 30: `\cr` (again)

```suggestion
#'    Otherwise, `join()` will throw an error message.
#'
#'    Example:
#'
```

R/qenv-join.R
```diff
@@ -7,102 +7,102 @@
...
-#' }
+#'    # Check the order of evaluation based on the id slot
+#'    shared_ids <- intersect(q@id, z@id)
+#'    match(shared_ids, q@id) # Output: 1 3
+#'    match(shared_ids, z@id) # Output: 1 2
+#'    ```
+#'    The error occurs because the index position of identical `@id` between the two objects is not the same.
+#'
+#' 3. The usage of temporary variable in the code expression could cause `join()` to fail.\cr\cr
+#'    Example:
```
averissimo, line 57: `\cr`

```suggestion
#' 3. The usage of temporary variable in the code expression could cause `join()` to fail.
#'
#'    Example:
#'
```

R/qenv-join.R
```diff
@@ -7,102 +7,102 @@
...
+#'      common_q,
+#'      "y <- numeric(0)
+#'       for (i in 1:3) {
+#'         y <- c(y, i)
+#'       }"
+#'    )
+#'    q <- join(x,y)
+#'    # Error message will occur
+#'
+#'    # Check the value of temporary variable i in both objects\cr
```
averissimo, line 77: `\cr` is appearing in the man page as it's inside a code block

```suggestion
#'    # Check the value of temporary variable i in both objects
```

averissimo (commented):
Good catch. Minor comments related to `\cr`.

- One `\cr` is being displayed in the man page. Please remove this.
- Consider replacing other `\cr` with markdown's 2 new lines, but very optional.

(Didn't mark as approved as auto-merge is on and documentation needs to be re-generated if suggestions are integrated)

R/qenv-join.R
```diff
@@ -7,102 +7,106 @@
...
-#' }
+#' 1. Both `qenv` objects contain an object of the same name but are not identical.
+#'
+#'    Example:
+#'
+#'    ```r
+#'    x <- qenv(
+#'      code = c(mtcars1 = "mtcars1 <- mtcars"),
+#'      env = list2env(list(mtcars1 = mtcars))
+#'    )
```
donyunardi, line 18: This code doesn't work for me:

image

Should this be updated using `within` or `eval_code` since we no longer recommend `new_qenv`?

donyunardi: Something like this?

```r
x <- qenv()
x <- eval_code(x, "mtcars1 <- mtcars")
y <- qenv()
y <- eval_code(y, "mtcars1 <- mtcars['wt']")
z <- join(x, y)
```

image

kartikeyakirar: These were not under @example tag, but were under documentation, so they got missed."

donyunardi (approved):
👍🏼

PR #207, opened 2024-08-23
206 fix failed pipeline

### LINKED ISSUE #206: Fix failed pipeline
https://github.com/pawelru/ie_status?tab=readme-ov-file

donyunardi:
Fixes #206 

Nosuggest pipeline passed:
https://github.com/insightsengineering/teal.code/actions/runs/10517334132/job/29141241618

### DIFF
--- a/vignettes/qenv.Rmd
+++ b/vignettes/qenv.Rmd
@@ -147,7 +147,7 @@ These functions can be seamlessly integrated into `shiny` applications to produc
 
 When employing a `qenv` to evaluate code, should an error occur, an object of type `qenv.error` is generated. This object can be utilized wherever a `qenv` object is used, alleviating the need for code alterations to handle these errors. Select the `error_option` in the example below to witness `qenv` error handling in action.
 
-```{r, eval=FALSE}
+```{r, eval=requireNamespace("shiny")}
 library(shiny)
 # create an initial qenv with the data in
 data_q <- qenv()
--- a/vignettes/qenv.Rmd
+++ b/vignettes/qenv.Rmd
@@ -35,7 +35,8 @@ print(my_qenv)
 get_env(my_qenv)
 
 
-q1 <- eval_code(my_qenv, "y <- x * 2") |> eval_code("z <- y * 2")
+q1 <- eval_code(my_qenv, "y <- x * 2")
+q1 <- eval_code(q1, "z <- y * 2")
 
 # my_qenv still contains only x
 print(my_qenv)
@@ -48,7 +49,8 @@ ls(get_env(q1))
 
 The same result can be achieved with the `within` method for the `qenv` class.
 ```{r}
-q2 <- within(my_qenv, y <- x * 2) |> within(z <- y * 2)
+q2 <- within(my_qenv, y <- x * 2)
+q2 <- within(q2, z <- y * 2)
 print(q2)
 ```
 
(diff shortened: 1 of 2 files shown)


donyunardi:

vignettes/qenv.Rmd
```diff
@@ -26,18 +26,16 @@ print(empty_qenv)
...
-library(magrittr)
-
 # evaluate code in qenv
 my_qenv <- eval_code(empty_qenv, "x <- 2")
 print(my_qenv)
 get_env(my_qenv)
 
 
-q1 <- eval_code(my_qenv, "y <- x * 2") %>% eval_code("z <- y * 2")
+q1 <- eval_code(my_qenv, "y <- x * 2") |> eval_code("z <- y * 2")
```
gogonzo, line 38: `|>` require R >= 4.1. I suggest to replace `|>` or `%>%` with normal nested R calls

vignettes/qenv.Rmd
```diff
@@ -147,11 +145,10 @@ These functions can be seamlessly integrated into `shiny` applications to produc
 
 When employing a `qenv` to evaluate code, should an error occur, an object of type `qenv.error` is generated. This object can be utilized wherever a `qenv` object is used, alleviating the need for code alterations to handle these errors. Select the `error_option` in the example below to witness `qenv` error handling in action.
 
-```{r}
+```{r, eval=FALSE}
```
pawelru, line 148: if it's about `shiny` you can do `eval = requireNamespace("shiny")` so that it would eval if one would call it interactively

donyunardi: @pawelru 

But isn't with `eval=FALSE`, you can still run the code interactively?
I don't see the point in evaluating this shiny code for vignette creation since there's no output anyway. 

I also don't see any difference in behavior between `eval=FALSE` and `eval=requireNamespace("shiny")` if the user doesn't have `shiny` installed and they run the chunk interactively. I tested this, and in both cases, an error occurs stating `Error in library(shiny): there is no package called 'shiny'`.

vignettes/qenv.Rmd
donyunardi: good idea, I will replace all `|>` occurences as you suggested.

vignettes/qenv.Rmd
pawelru: Sorry - in my previous message I meant "if one has `shiny` preinstalled" instead of "inside interactive run". 

You are right about running the code by the end user. Regardless of the `eval` parameter, code chunks can be freerly copied and evaluated inside user session.

Package vignettes needs to be built, then installed and only then these pre-built files can be browsed by the end user. It's quite likely that `shiny` will be there during the build process hence this code chunk would be actually evaluated. We do want to have as much code auto-tested as possible. I'm afraid that with `eval = FALSE` we are going to hard-exclude too much.

donyunardi: > We do want to have as much code auto-tested as possible.

Yes, I agree with this. 👍🏼

vignettes/qenv.Rmd
```diff
@@ -26,18 +26,17 @@ print(empty_qenv)
 
 ### `qenv` basic usage
 
-The `eval_code()` function executes code within a `qenv` environment, yielding a new `qenv` object as the output. 
+The `eval_code()` function executes code within a `qenv` environment, yielding a new `qenv` object as the output.
 
 ```{r}
-library(magrittr)
```
pawelru, line 32: I can see that you have successfully removed this package. In case this is the only usage then we can remove it from the suggest. I'm on the phone so can't check it now lol 😉

pawelru (approved):

vignettes/qenv.Rmd
donyunardi: There is traces of `%>%` in unit tests so I don't think we can't remove this until we rewrite those test cases.

Issue #212, opened 2024-10-15
[Feature Request]: Code to store and return comments


gogonzo:
### Feature description

Code should accept and return comments. An obstacle is the fact that we turn code to language at some point so we are loosing comments.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

Issue #211, opened 2024-09-26
[Feature Request]: "subset" method for `qenv` object


gogonzo:
### Feature description

`subset(names)` method, should subset `qenv` object to keep only `@env <- @env[names]` and `@code <- <@code for [names]>`.
Consider if it is worth exposing such function, or should we keep it as `teal` internal (where it is used).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

donyunardi:
### Acceptance Criteria
* Decide on the name for the new method.
* Create a new method for `qenv` in teal.code
* Ensure to also able to subset `join_keys` in teal.data

m7pr:
Hey, I think I can start on this one, as this is one of the blockers for teal release. I think we should start with teal.code and teal.data cards as soon as possible, so that we can unblock teal release

PR #139, opened 2023-08-23
133 code parser

m7pr:
Closes #133 

This PR introduces a feature that can be utilized in a broad usage. Currently it only extends `qenv` class, but the big picture is that we will be able to change the way we provide `data` in `teal::init`.

# Current behavior
Currently `teal::init` takes `teal.data::teal_data()` as an input which takes actual R objects as an input with an extra companion of code specification, which in all cases is the code used to create the R object that is being passed. This results in code duplication: we first create a code to create the object, and then we copy-paste this code into an object specification at `teal.data::teal_data()`. 

```{R}
new_iris <- transform(iris, id = seq_len(nrow(iris)))
new_mtcars <- transform(mtcars, id = seq_len(nrow(mtcars)))

app <- init(
  data = teal_data(
    dataset("new_iris", new_iris, code = "new_iris <- transform(iris, id = seq_len(nrow(iris))"),
    dataset("new_mtcars", new_mtcars, code = "new_mtcars <- transform(mtcars, id = seq_len(nrow(mtcars)))"),
  )
)
```

# Proposed alternative
The alternative to the above, proposed in this PR, is having a functionality called code parser. This functionality understands which parts of the code (passed as a character) is needed to create a specific object (with all it's dependent objects and dependent side-effects). Thanks to that, we don't need to pass object specifications and code separately - we can just pass the code, which will be evaluated and which will be parsed so that under the hood objects are created and their respective code is assigned to them automatically.

This PR only introduces changes to `qenv` object. Further changes to how `teal.data::teal_data()` or `teal::init` `data` parameter work will be needed. `qenv` received a new field called `code_dependency` that is a list needed to restore the object and it's side effects. Below are a few examples on extraction of the code of `ADSL` object

<details>
<summary>code objects </summary>

```r
library(dplyr)
code = '
arm_mapping <- list(
  "A: Drug X" = "150mg QD",
  "B: Placebo" = "Placebo",
  "C: Combination" = "Combination"
)
color_manual <- c("150mg QD" = "#000000", "Placebo" = "#3498DB", "Combination" = "#E74C3C")
# assign LOQ flag symbols: circles for "N" and triangles for "Y", squares for "NA"
shape_manual <- c("N" = 1, "Y" = 2, "NA" = 0)
ADSL <- goshawk::rADSL
goshawk::rADLB-> ADLB
iris2 <- iris # @effect ADLB ADSL
var_labels <- lapply(ADLB, function(x) attributes(x)$label)
iris3 <- iris'
code2 = '
[...]

PR #216, opened 2024-10-23
211 `[.qenv` S3 method + replacement of `@id`, `@warnings`, and `@messages` fields

m7pr:
Closes

- #211 
- #70 
- #212 

Changes

- Modifies slot design for `@code`, `@id`, `@messages`, `@warning` so they are all part of the `@code` attributes
- Introducing `[.` method for `qenv` that reduces the object to the `code` and `objects` limited to those provided in `names` parameter. 
- Changed the way `eval_code` and `get_code` analyze the code. 
    - They analyze the code by single call right now.
    - Code is kept as is, without any changes in the indentation.
    - Code dependency is calculated during `eval_code`.

Checklist:

- [x] decide what to do with `@id`, `@warnings`, `@messages` fields
- [x]  create `[.` in `teal_data` package so it also handles `join_keys`, `@verified` and potentially `datanames` https://github.com/insightsengineering/teal.data/pull/346
- [x] create tests for subset
- [x] fix tests
- [x] invent a new name since `subset` is an S3 method in `base` package
- [x] check if tests in `teal.data` are not failing after this change
- [x] check if tests in `teal` are not failing after this change https://github.com/insightsengineering/teal/pull/1397
- [ ] check if tests in `tmc` and `tmg` are not failing after this change

Tested with

```r
library(teal.code)

# example 1 -------------------------------------------------------------------------------------------------------

code <- "
  # initial comment
  a <- 1
  b <- 2 # inline comment
  c <- 3
  # inbetween comment
  d <- 4
  # finishing comment 
"
q <- qenv() |> eval_code(code)
q@code
q |> get_code()
q |> get_code(names = "c")
q['c']
q['c']@code
q['x']
q['x']@code


# example 2 -------------------------------------------------------------------------------------------------------

code <- "
  # initial comment line 1
  # initial comment line 2
  a <- 1 # A comment
  b1 <- 2; b2 <- 2;b3 = 3 # inline comment
  c <- 3 # C comment
  # inbetween comment
  d <- 4
  # finishing comment line 1
  # finishing comment line 2
"

q <- qenv() |> eval_code(code)
q@code
q |> get_code()
q |> get_code(names = c('a', 'd'))
q['a']
q['a']@code
q[c('a', 'b1', 'd')]
q[c('a', 'b1', 'd')]@code
q[c('a', 'BULBASAUR', 'd')]
q[c('a', 'BULBASAUR', 'd')]@code


# example 3 -------------------------------------------------------------------------------------------------------

code <- "
  # initial comment line 1
  # initial comment line 2
  o1 <- 1; o2 <- 2; o3 <- 3
  a <- 1 # A comment
  b1 <- 2; b2 <- 2;b3 = 3 # inline comment
  c <- 3 # C comment
  # inbetween comment
  d <- 4
  # finishing comment line 1
  # finishing comment line 2
"

q <- qenv() |> eval_code(code)
q@code
q |> get_code()
q |> get_code(names = c('a', 'o2', 'd'))
q['a']
q['a']@code
q[c('a', 'b1', 'd')]
q[c('a', 'b1', 'd')]@code
q[c('a', 'b1', 'd', 'o2')]
q[c('a', 'b1', 'd', 'o2')]@code

```


R/qenv-subset.R
```diff
@@ -0,0 +1,52 @@
...
+  new_qenv@env <- list2env(mget(x = names_in_env, envir = get_env(object)))
+  new_qenv@code <- get_code(object, names = names_in_env)
+  # Question: what about @id, @warnings, @messages?
+  # Currently:
+  # > new_qenv@id
+  # integer(0)
+  # > new_qenv@warnings
+  # character(0)
+  # > new_qenv@messages
+  # character(0)
```
m7pr, line 43: The implementation of `.subset_teal_data` in teal omits those fields and they are empty
https://github.com/insightsengineering/teal/blob/a1087d2d3ff0c62393c3d5277cd5f184d543e2d9/R/teal_data_utils.R#L41-L64

Should we do the same in here?

m7pr:
@gogonzo what do you think about creating a `subset` in `teal.code` that handles `@code` and `@env` (basics of `qenv` class) and another `subset` method in `teal.data` that handles `join_keys`, `@verified` and potentially `datanames` (major points of `teal_data` class)?

gogonzo:
> @gogonzo what do you think about creating a `subset` in `teal.code` that handles `@code` and `@env` (basics of `qenv` class) and another `subset` method in `teal.data` that handles `join_keys`, `@verified` and potentially `datanames` (major points of `teal_data` class)?

Theoretically if you don't make a new `qenv` object but only replace `@env` and `@code` fields then you don't need to make extra changes for `teal.data` because its slots will be preserved. 


>  decide what to do with @id, @warnings, @messages fields

You already noticed a problem with `@id`, `@warnings` and `@messages` - they probably should be moved to attributes of each `@code` element, so when you subset @code - they will be automatically subset as well.

gogonzo:
I see the problem. `@code` element might contain more than one R call. Maybe we should split `code` which enters `eval_code` and keep a `@code` per call. Then it would be straightforward to match indexes obtained in `get_code_dependency` with `@code`.

m7pr:
Alrighty, that would make sense. Should we do the same for how `within` assigns the `@code` to the resulted `qenv`?

NAMESPACE
```diff
@@ -13,6 +13,7 @@ export(get_warnings)
 export(join)
 export(new_qenv)
 export(qenv)
+export(subset)
```
m7pr, line 16: We need to come up with a different name, since `subset` is an S3 function in `base`

```r
> library(teal.code)

Attaching package: ‘teal.code’

The following object is masked from ‘package:base’:

    subset
```

R/qenv-eval_code.R
```diff
@@ -28,18 +28,20 @@
 setGeneric("eval_code", function(object, code) standardGeneric("eval_code"))
 
 setMethod("eval_code", signature = c("qenv", "character"), function(object, code) {
-  id <- sample.int(.Machine$integer.max, size = 1)
+  parsed_code <- parse(text = code, keep.source = TRUE)
+  comments <- extract_comments(parsed_code)
+  id <- sample.int(.Machine$integer.max, size = length(parsed_code))
```
m7pr, line 33: now all `@id`, `@code`, `@warnings`, `@messages` will have the same length as number of single calls in input `code`. The resulted object will have longer fields (especially code field). We need to double check if that doesn't brake `teal` and `teal.data` and Show R Code in modules.

m7pr:
@gogonzo I pushed all the changes I planned for this PR. There are 4 tests failing but I think it's proper time to review the current state. We need to double check if those changes do not brake any tests in `teal.data` and `teal`, `tmc/tmg`. I will take care of that

R/qenv-get_code.R
```diff
@@ -124,11 +124,12 @@ setMethod("get_code", signature = "qenv", function(object, deparse = TRUE, names
...
-    if (length(code) == 0) {
-      code
-    } else {
-      paste(code, collapse = "\n")
-    }
+    # if (length(code) == 0) {
+    #   code
+    # } else {
+    #   paste(code, collapse = "\n")
+    # }
```
m7pr, line 131:
```suggestion
```

tests/testthat/test-qenv_within.R
```diff
@@ -79,28 +79,28 @@ testthat::test_that("external values can be injected into expressions through `.
 
   external_value <- "virginica"
   q <- within(q, {
-    i <- subset(iris, Species == species)
+    i <- base::subset(iris, Species == species)
   },
   species = external_value)
 
-  testthat::expect_identical(get_code(q), "i <- subset(iris, Species == \"virginica\")")
+  testthat::expect_identical(get_code(q), "i <- base::subset(iris, Species == \"virginica\")")
```
m7pr, line 86: this will be reverted, once we change the name of the `subset` method to something else. right not it clashes

tests/testthat/test-qenv_within.R
```diff
@@ -79,28 +79,28 @@ testthat::test_that("external values can be injected into expressions through `.
...
 })
 
 testthat::test_that("external values are not taken from calling frame", {
   q <- qenv()
   species <- "setosa"
   qq <- within(q, {
-    i <- subset(iris, Species == species)
+    i <- base::subset(iris, Species == species)
```
m7pr, line 93: this will be reverted, once we change the name of the `subset` method to something else. right not it clashes

tests/testthat/test-qenv_within.R
```diff
@@ -79,28 +79,28 @@ testthat::test_that("external values can be injected into expressions through `.
...
   })
   testthat::expect_s3_class(qq, "qenv.error")
   testthat::expect_error(get_code(qq), "object 'species' not found")
 
   qq <- within(q, {
-    i <- subset(iris, Species == species)
+    i <- base::subset(iris, Species == species)
```
m7pr, line 99: this will be reverted, once we change the name of the `subset` method to something else. right not it clashes

tests/testthat/test-qenv_within.R
```diff
@@ -79,28 +79,28 @@ testthat::test_that("external values can be injected into expressions through `.
...
   },
   species = species)
   testthat::expect_s4_class(qq, "qenv")
-  testthat::expect_identical(get_code(qq), "i <- subset(iris, Species == \"setosa\")")
+  testthat::expect_identical(get_code(qq), "i <- base::subset(iris, Species == \"setosa\")")
```
m7pr, line 103: this will be reverted, once we change the name of the `subset` method to something else. right not it clashes

NAMESPACE
m7pr: propositions for new name

- trim
- limit
- filter
- reduce
- shrink
- crop
- select
- slice
- truncate
- prune

gogonzo: If you want S3 method for qenv you just need to make S3 methods `subset.qenv <- `, `subset.qenv.error <-`. Since we are talking about different names, what is the most relevant name for this? What would you like to use to trim qenv object?

m7pr: I would keep all the methods as S4 so it's consistent across other methods that we have for `qenv`.
I think `subset`, `truncate`, `trim` of `filter` are the most relevant in here.

gogonzo: Let's consider how we subset slots:
- we subset `@env` with `@env[<names>]`
- we subset `@join_keys` with `@env[<names>]`

`[.qenv` also makes sense IMO

m7pr: `@env` is not subsettable

```r
> x <- new.env(); x$a <- 1; x$b <- 2
> x[c('a', 'b')]
Error in x[c("a", "b")] : object of type 'environment' is not subsettable
> x[[c('a', 'b')]]
Error in x[[c("a", "b")]] : wrong arguments for subsetting an environment
```

`[.` works only for names of length 1, that's why we use `mget` in the implementation.

Because of this reason I would strive from `[.`, since environments are not subsettable

gogonzo: @m7pr How would you address subsetting of `teal_data` object in the future? Specifically `join_keys` slot?

m7pr: @gogonzo `join_keys` is just an S3 list with custom `attributes` and a customized `print` method. So using `[.` or `[[.` is natural in this case.

`teal_data` and `qenv` are a bit more complex objects as they have multiple fields of different classes.
`teal_data` doesn't handle `[[.` for multiple names, just for one and returns the dataset. `teal_code` does the same with `[[.` being a synonym to `get_var`


```r
data <- teal_data(iris = iris, mtcars = mtcars)
data[['iris']] # works 
data[[c('iris', 'mtcars')]] # doesn't work
first argument has length > 1
```

Both `teal_data` and `qenv` already have defined methods for `[[.` that do `get_var`

```r
> getMethod("[[","qenv")
Method Definition:

function (x, i, j, ...) 
{
    .local <- function (x, i) 
    {
        get_var(x, i)
    }
    .local(x, i, ...)
}
<bytecode: 0x000001cff0ac0f78>
<environment: namespace:teal.code>

Signatures:
        x     
target  "qenv"
defined "qenv"
```


So the idea is to introduce a method with a different name than `[[.`, for example `truncate` that returns trimmed `qenv` and respected `@code`, `@env`, `@warnings` and `@messages` fields. For `teal_data` we can introduce the same method with the same name that also handles `@verified`, `@datanames` and `@join_keys` fields if needed.

R/utils-get_code_dependency.R
```diff
@@ -64,7 +65,8 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
 
   lib_ind <- detect_libraries(calls_pd)
 
-  as.character(code[sort(unique(c(lib_ind, ind)))])
+  code_ids <- sort(unique(c(lib_ind, ind)))
+  trimws(paste(as.character(code[code_ids]), comments[code_ids]))
```
m7pr, line 69: `teal.code::get_code_dependency` now returns comments as well, so that when it is used in `subset` it creates an object that does contain comments in the final `@code`.

We might edit `teal.code::get_code` so it drops comments, so this will not be shown in the final extracted code.
Check this PR that shows what is changed in the returned code https://github.com/insightsengineering/teal/pull/1396

Comments are included.

R/qenv-subset.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @name subset
+#' @rdname qenv
+#' @aliases subset,qenv-method
+#' @aliases subset,qenv.error,ANY-method
+#'
+#' @export
+setGeneric("subset", function(object, names) standardGeneric("subset"))
+
+setMethod("subset", signature = c("qenv"), function(object, names) {
```
m7pr, line 25: based on 

https://github.com/insightsengineering/teal/blob/a1087d2d3ff0c62393c3d5277cd5f184d543e2d9/R/teal_data_utils.R#L41-L64

tests/testthat/test-qenv_get_warnings.R
```diff
@@ -59,7 +57,7 @@ testthat::test_that("get_warnings accepts a qenv object with 1 warning eval_code
     eval_code(bquote("x <- 1")) %>%
     eval_code(bquote(warning("This is a warning 2!")))
   testthat::expect_identical(
-    get_warnings(q),
+    get_warnings(q), # TODO fix
```
m7pr, line 60:
```suggestion
    get_warnings(q),
```

tests/testthat/test-qenv_get_warnings.R
```diff
@@ -45,11 +45,9 @@ testthat::test_that("get_warnings accepts a qenv object with a single eval_code
...
     get_warnings(q),
     paste0(
-      "~~~ Warnings ~~~\n\n",
-      "> This is a warning 1!\n> This is a warning 2!\nwhen running code:\n",
-      "warning(\"This is a warning 1!\")\nwarning(\"This is a warning 2!\")\n\n",
-      "~~~ Trace ~~~\n\n",
-      "warning(\"This is a warning 1!\")\nwarning(\"This is a warning 2!\")"
+      "~~~ Warnings ~~~\n\n> This is a warning 1!\nwhen running code:\nwarning(\"This is a warning 1!\")",
+      "\n\n> This is a warning 2!\nwhen running code:\nwarning(\"This is a warning 2!\")\n\n",
+      "~~~ Trace ~~~\n\nwarning(\"This is a warning 1!\")\nwarning(\"This is a warning 2!\")"
```
m7pr, line 50:
```suggestion
      "~~~ Warnings ~~~\n\n",
      "> This is a warning 1!\n> This is a warning 2!\nwhen running code:\n",
      "warning(\"This is a warning 1!\")\nwarning(\"This is a warning 2!\")\n\n",
      "~~~ Trace ~~~\n\n",
      "warning(\"This is a warning 1!\")\nwarning(\"This is a warning 2!\")"
```

NAMESPACE
gogonzo: > join_keys is just an S3 list with custom attributes and a customized print method. So using [. or [[. is natural in this case.

What "natural" means? There is `[.join_keys` overriden in the `teal.data` so it is not entirely natural.

> teal_data and qenv are a bit more complex objects as they have multiple fields of different classes.
teal_data doesn't handle [[. for multiple names, just for one and returns the dataset. teal_code does the same with [[. being a synonym to get_var

`[[` doesn't handle multiple index in general. For example `list("a", "b")[[c(1, 2)]]` fails. In general `[` and `[[` are completely different functions - `[` subsets and keeps the original class while `[[` extracts element.  I suggested possible usage of `[` for `qenv` which in certain design wouldn't require introducing extra `[.teal_data`.

m7pr: Alrighty @gogonzo , so `[[.` exists for `qenv` and extracts one element. `[.` does now exist yest for `qenv` but could exist and could extract multiple elements. I think it make sense. Thanks for the clarification. I will continue with the proposition to use `[.`

NAMESPACE
```diff
@@ -1,5 +1,6 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method("[",qenv)
```
m7pr, line 3: Any way to have this method be documented as the last one in the documentation, not the first one?
image

gogonzo: please see `within.qenv`. `within` is listed in [package index](https://insightsengineering.github.io/teal.code/latest-tag/reference/qenv.html)

m7pr: I think I didn't rebuild the documentation locally. This is how it looks, and it's proper now

The order is taken by the alphabetical order of R-files

- qenv -> qenv-constructor.R
- new_qenv -> qenv-constructor.R
- eval_code -> qenv-eval_code.R
- x[datanames] -> qenv-extract.R
- get_code -> qenv-get_codeR
- within -> qenv-withinr.R

image

R/qenv-eval_code.R
```diff
@@ -28,18 +28,20 @@
...
 
   object@id <- c(object@id, id)
   object@env <- rlang::env_clone(object@env, parent = parent.env(.GlobalEnv))
-  code <- paste(code, collapse = "\n")
-  object@code <- c(object@code, code)
+  object@code <- c(object@code, trimws(paste(as.character(parsed_code), comments)))
```
gogonzo, line 37: comments should be extracted per single call. This makes `get_code` to return wrong code:

```r

x <- qenv() |> eval_code("
  # initial comment
  a <- 1
  b <- 2 # inline comment
  c <- 3
  # inbetween comment
  d <- 4
  # finishing comment 
")

get_code(x)
# [1] "a <- 1 # initial comment"   "b <- 2 # inline comment"   
# [3] "c <- 3"                     "d <- 4 # inbetween comment"
# [5] "a <- 1 # finishing comment"
```

R/qenv-eval_code.R
```diff
@@ -28,18 +28,20 @@
...
-  current_warnings <- ""
-  current_messages <- ""
+  current_warnings <- rep("", length(parsed_code))
+  current_messages <- rep("", length(parsed_code))
 
-  parsed_code <- parse(text = code, keep.source = TRUE)
-  for (single_call in parsed_code) {
+
+  for (i in seq_along(parsed_code)) {
+    single_call <- parsed_code[i]
```
gogonzo, line 44: Parsing independently calls and comments is a risky business. Also we distort original `code` which is important for a [future issue](https://github.com/insightsengineering/teal.code/issues/212).

```suggestion
  code_split <- split_code(code) # keep original character
  for (i in seq_along(code_split)) {
    single_call <- code_split[[i]]
    ...
```

gogonzo: Also co-ocurrence and side-effects can be extracted when evaluating attach this information to each call.
Then `get_code_dependencies` will be simplified

m7pr:
Thanks for the review @gogonzo - very useful feedback

R/qenv-eval_code.R
gogonzo: `split_code` could split into calls, but @code needs to be stored with comments. If you use parse and `utils::getParsedData` then you'll need to convert back to the character. I think using `getParsedData` will be the most accurate to determine the end of the call.

R/qenv-eval_code.R
m7pr: @gogonzo thanks for finding that! It looks like this was caused by the inability of `utils::getParsedCode` to extract comments properly. We had `fix_shifted_comments` function that was used in `get_code_dependency`. I reused that function in `eval_code` to match comments with the calls. Thanks! also added 2 new tests

38bd16fb25b6a3a27d473d41e5694155cb29ccbf

```r
  # If comment is on top, it gets moved to the first call.
  # Any other comment gets moved to the call above.
  # Comments get pasted if there are two assigned to the same call.
library(teal.code)
code <- "
    # initial comment
    a <- 1 # A comment
    b <- 2 # inline comment
    c <- 3 # C comment
    # inbetween comment
    d <- 4
    # finishing comment
  "

qenv() |> eval_code(code) |> get_code()
#> [1] "a <- 1 # initial comment # A comment"  
#> [2] "b <- 2 # inline comment"               
#> [3] "c <- 3 # C comment # inbetween comment"
#> [4] "d <- 4 # finishing comment"
```

<sup>Created on 2024-10-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/qenv-eval_code.R
m7pr: I think we can work on a simplified scenario. Let's just limit ourself to the current input and functions.

```r
code <- "
  # initial comment
  a <- 1 # A comment
  b <- 2 # inline comment
  c <- 3 # C comment
  # inbetween comment
  d <- 4
  # finishing comment
"

parsed_code <- parse(text = code)
comments <- extract_comments(parsed_code)

for(single_call in parsed_code) eval(single_call)

trimws(paste(as.character(parsed_code), comments))
```

Pay attention that
- `eval` inside `for` loop works on `expressions`
- we do not have a way to divide `code` into a calls in another way than through the inbuild parser or `parse(text = `
- we are able to `extract_comments` in our way, even thought sometimes there is more comments than calls

> # keep original character

I don't see a way how we can apply `eval` on `character` and how to divide character `code` into calls without parsing.

R/qenv-eval_code.R
m7pr: Based on our conversation I prepared a (dirty prototype) function that divides the input code by calls and keeps comments as they were in the initial character input. 56b5c8e824383a4762c53f0dc78b566888eabeda

We need to decide if we want to pull the initial comments to the first call, and the final comments to the last call, as they are currently extracted as individual calls. Also the inline comment gets taken into the previous calls, not to the next call. This is also something that we can discuss.


```r
code <- "
  # initial comment line 1
  # initial comment line 2
  a <- 1 # A comment
  b1 <- 2; b2 <- 2;b3 = 3 # inline comment
  c <- 3 # C comment
  # inbetween comment
  d <- 4
  # finishing comment line 1
  # finishing comment line 2
"

parsed_code <- parse(text = code)
comments <- extract_comments(parsed_code)
pd <- utils::getParseData(parsed_code)

pd <- pd[pd$token != "';'", ]
lines_ids <- get_line_ids(pd)
code_by_calls <- split_code(code, lines_ids)
code_by_calls
```


```r
[[1]]
[1] "# initial comment line 1\n  # initial comment line 2"

[[2]]
[1] "a <- 1 # A comment"

[[3]]
[1] "b1 <- 2"

[[4]]
[1] "b2 <- 2"

[[5]]
[1] "b3 = 3 # inline comment"

[[6]]
[1] "c <- 3 # C comment\n  # inbetween comment"

[[7]]
[1] "d <- 4"

[[8]]
[1] "# finishing comment line 1\n  # finishing comment line 2"
```

m7pr:
Hey, this is the state where I left it for today

``` r
library(teal.code)

code <- "
  # initial comment
  a <- 1
  b <- 2 # inline comment
  c <- 3
  # inbetween comment
  d <- 4
  # finishing comment 
"

qenv() |> eval_code(code) |> get_code()
#> [1] "# initial comment"             "a <- 1"                       
#> [3] "b <- 2 # inline comment"       "c <- 3\n  # inbetween comment"
#> [5] "d <- 4"                        "# finishing comment "

code <- "
  # initial comment line 1
  # initial comment line 2
  a <- 1 # A comment
  b1 <- 2; b2 <- 2;b3 = 3 # inline comment
  c <- 3 # C comment
  # inbetween comment
  d <- 4
  # finishing comment line 1
  # finishing comment line 2
"

qenv() |> eval_code(code) |> get_code()
#> [1] "# initial comment line 1\n  # initial comment line 2"    
#> [2] "a <- 1 # A comment"                                      
#> [3] "b1 <- 2"                                                 
#> [4] "b2 <- 2"                                                 
#> [5] "b3 = 3 # inline comment"                                 
#> [6] "c <- 3 # C comment\n  # inbetween comment"               
#> [7] "d <- 4"                                                  
#> [8] "# finishing comment line 1\n  # finishing comment line 2"
```

<sup>Created on 2024-10-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

m7pr:
But something is off with `get_code_dependency`. Do not review yet, I'll continue tomorrow

m7pr:
Do not review yet, there are tests failing still

m7pr:
Hey @gogonzo thanks for the time you invested in giving feedback. I think I was able to create `split_code` function that satisfies our needs, and also handles the edge cases we listed in tests. I updated the initial comment in this PR so it included the list of changes, and also the code I used for testing https://github.com/insightsengineering/teal.code/pull/216#issue-2608047134

The biggest change is that the code is analyzed by single calls, and we keep comments and we do not change the formatting/indentation of the code anymore. 

There are still 2 tests failing, one is minor, but the one in `within` sounds a bit bigger.

tests/testthat/test-qenv_within.R
```diff
@@ -60,6 +60,8 @@ testthat::test_that("within.qenv renturns a `qenv` where `@env` is a deep copy o
   qq <- within(q, {})
   testthat::expect_equal(q@env, qq@env)
   testthat::expect_false(identical(q@env, qq@env))
+  # TODO: fix
+  # dunno what's going on yet
```
gogonzo, line 64: In eval_code you've introduced following line which exists before environment is cloned.
```r
  if (length(parsed_code) == 0) {
    return(object)
  }
```

gogonzo (commented):
I believe that workflow with code, id, warnings, messages can be simplified. I need more time to analyse `split_code`

R/qenv-extract.R
```diff
@@ -0,0 +1,48 @@
+#'
+#' @section Subsetting:
+#' `x[names]` subsets objects in `qenv` environment and limit the code to the necessary needed to build limited objects.
+#'
+#' @param names (`character`) names of objects included in `qenv` to subset
```
pawelru, line 5: please mention about warnings

pawelru, line 5: in the Rd file you are writing (i.e. `@rdname qenv` -> rendered into https://insightsengineering.github.io/teal.code/main/reference/qenv.html) there is already a parameter `names` with a different description. Have you tried how this renders to a single rd file? Would it be confusing?

R/qenv-extract.R
```diff
@@ -0,0 +1,37 @@
...
+`[.qenv` <- function(x, names) {
+  checkmate::assert_class(names, "character")
+  names_in_env <- intersect(names, ls(get_env(x)))
+  if (!length(names_in_env)) {
+    warning("None of `names` elements exist in `qenv`. Returning empty `qenv`.")
+    return(qenv())
+  }
+
+  limited_code <- get_code(x, names = names_in_env)
+  indexes <- which(x@code %in% limited_code)
```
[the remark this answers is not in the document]
m7pr: I don't think we should remove duplicates. Imagine a function called `update_num()` that, when applied on a numeric object, updates the object by 1. So most of the calls would look the same, but the change the object so they are needed.

``` r
update_num <- function(name) {
  x <- get(name, envir = .GlobalEnv)
  x <- x + 1
  assign(x = name, value = x, envir = .GlobalEnv)
}

z <- 1
update_num("z")
z
#> [1] 2
update_num("z")
z
#> [1] 3
update_num("z")
z
#> [1] 4
```

<sup>Created on 2024-10-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/qenv-eval_code.R
```diff
@@ -28,18 +28,24 @@
...
+  code_split <- split_code(paste(code, collapse = "\n"))
+  object@code <- c(object@code, unlist(code_split))
 
-  parsed_code <- parse(text = code, keep.source = TRUE)
-  for (single_call in parsed_code) {
+  current_warnings <- rep("", length(parsed_code))
+  current_messages <- rep("", length(parsed_code))
+
+  for (i in seq_along(code_split)) {
+    single_call <- parse(text = code_split[[i]], keep.source = FALSE)
```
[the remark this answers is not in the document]
m7pr: @gogonzo we've been before with `occurrence` and `side_effects` calculation inside `eval_code`. That was the first attempt for the code parser. It extended `qenv` class and kept `side_effects` and `occurrence` updated after every `eval_code`. We decided to keep `eval_code` to just evaluate the code, and use `get_code` to analyze the code. That was the initial PR https://github.com/insightsengineering/teal.code/pull/139

We also already had `get_code_dependency` in `teal.code` but removed it in this PR https://github.com/insightsengineering/teal.code/pull/167

R/qenv-extract.R
```diff
@@ -0,0 +1,37 @@
...
+
+  x@env <- list2env(mget(x = names_in_env, envir = get_env(x)))
+  x@code <- limited_code
+  x@id <- x@id[indexes]
+  x@warnings <- x@warnings[indexes]
+  x@messages <- x@messages[indexes]
```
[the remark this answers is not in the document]
m7pr: I'm totally fine with it. Easier to trim the `qenv` object. You just limit by `@env` names, and then by `@code`. 
Each element of `@code` would have 3 attributes: `id`, `warnings`, `messages`.

Should we keep that in the scope of this PR, or start a new issue? For the planning and sprint scoping I would suggest to start a new issue for this.

tests/testthat/test-qenv_within.R
m7pr: Ah, thanks for the info!

R/qenv-extract.R
m7pr: Good idea! Will mention that some objects are skipped if not present, but the warning is raised!

m7pr: Oh yes indeed. Good catch. Will try to think about it, and maybe just document in `qenv.R` (both meanings) and skip in here.

pawelru: that might be a good idea. Then the existing description would need to be generalised to be true for the methods we have there

R/qenv-get_warnings.R
```diff
@@ -49,10 +49,9 @@ setMethod("get_warnings", signature = c("qenv"), function(object) {
   )
   lines <- Filter(Negate(is.null), lines)
 
-  sprintf(
-    "~~~ Warnings ~~~\n\n%s\n\n~~~ Trace ~~~\n\n%s",
-    paste(lines, collapse = "\n\n"),
-    get_code(object)
+  paste0(
+    sprintf("~~~ Warnings ~~~\n\n%s\n\n", paste(lines, collapse = "\n\n")),
+    sprintf("~~~ Trace ~~~\n\n%s", paste(get_code(object), collapse = "\n"))
```
pawelru, line 54:
```suggestion
  paste0(
    sprintf("~~~ Warnings ~~~\n\n%s", paste(lines, collapse = "\n\n")),
    sprintf("~~~ Trace ~~~\n\n%s", paste(get_code(object), collapse = "\n")),
    collapse = "\n\n"
```

R/qenv-extract.R
```diff
@@ -0,0 +1,48 @@
+#'
+#' @section Subsetting:
+#' `x[names]` subsets objects in `qenv` environment and limit the code to the necessary needed to build limited objects.
+#'
+#' @param names (`character`) names of objects included in `qenv` to subset
+#' @param x (`qenv`)
+#'
+#' @examples
+#'
```
averissimo, line 9: Would it make sense to expand the subset to include all objects that are related (by code dependency)

See pseudo-test below.

This creates an inconsistency where the code and environment doesn't match. ([issue in teal](https://github.com/insightsengineering/teal/issues/1390))

``` r
pkgload::load_all("/teal.code")
#> ℹ Loading teal.code
q <- qenv() |> within({
  a <- 2
  b <- a + 1
})

q2 <- q["b"]

get_code(q2) |> paste(collapse = "\n") |> cat()
#> a <- 2
#> b <- a + 1
expect_identical(get_code(q2), c("a <- 2", "b <- a + 1"))

ls(get_env(q2))
#> [1] "b"
expect_setequal(ls(get_env(q2)), c("a", "b"))
#> Error: ls(get_env(q2)) (`actual`) and c("a", "b") (`expected`) don't have the same values.
#> * Only in `actual`: "a"
```

<sup>Created on 2024-10-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

m7pr: Oh yeah, that's very very needed

m7pr: @averissimo you would probably not like to extract `a` in case `a` is modified after `b`

```r
q <- qenv() |> within({
  a <- 2
  b <- a + 1
  a <- a + 2
})
```

What is the point of having `a` extracted in `@env` for `q["b"]` in this case? 
We just need `b` and the code needed to reproduce `b`.

averissimo: That's a really good point and my only thought is to simplify, instead of adding complexity.

We could "cheat" by removing non-subsetted objects, let's say in pseudo-code

``` r
q <- qenv() |> within({
  a <- 2
  b <- a + 1
  a <- a + 2
})

q2 <- q["b"]

q2 |> get_code() |> cat()
#> a <- 2
#> b <- a + 1
#> a <- a + 2
#> rm(a)
```


Subsetting the objects to get all related dependencies sounds good on paper in general, and it makes sense when considering `qenv` in isolation.

Upstream on teal, the modules must be specific and be included on the `datanames` for it to be used on the modules. So that issue may be closed.

R/qenv-get_code.R
```diff
@@ -118,19 +118,16 @@ setMethod("get_code", signature = "qenv", function(object, deparse = TRUE, names
...
     object@code
   }
 
   if (deparse) {
-    if (length(code) == 0) {
-      code
-    } else {
-      paste(code, collapse = "\n")
-    }
+    code
```
m7pr, line 128: @gogonzo 
We don't have `unlist(code)` in here, so you do not loose attributes of `code` in `[.qenv` when you subset the code.
This means in all places you need to run `unlist(get_code())` to get code represented as a character, not as a list.

R/qenv-get_code.R
```diff
@@ -118,19 +118,16 @@ setMethod("get_code", signature = "qenv", function(object, deparse = TRUE, names
   }
 
   code <- if (!is.null(names)) {
+    # todo:
```
m7pr, line 121:
```suggestion
```

R/qenv-get_code.R
m7pr: or we could steer the behavior with a parameter, so it unlist for get_code but does not unlist for `[.`

m7pr:
Hey, so far so good. We just need to edit tests so that such occurrences

```r
testthat::expect_identical(
  get_code(q, names = "a"),
  code
)
```

are changed to 

```r
testthat::expect_identical(
  paste(get_code(q, names = "a"), collapse = ""),
  paste(code, collapse = "\n")
)
```

R/utils-get_code_dependency.R
```diff
@@ -451,3 +452,61 @@ normalize_pd <- function(pd) {
...
+#' @return list of `character`s of the length equal to the number of calls in `code`.
+#'
+#' @keywords internal
+#' @noRd
+split_code <- function(code) {
+  call_breaks <- get_call_breaks(code)
+  if (nrow(call_breaks) == 1) {
+    return(code)
+  }
+  call_breaks <- call_breaks[order(call_breaks[, "line"], call_breaks[, "col"]), ]
```
m7pr, line 494: In case this has one row it gets changed to integer and looses `matrix` class
```suggestion
  call_breaks <- call_breaks[order(call_breaks[, "line"], call_breaks[, "col"]), , drop = TRUE]
```

m7pr:
I did change a couple of tests but few (around 5-10) are still failing. In some cases the `# @linksto` tag got messed up

R/utils-get_code_dependency.R
```diff
@@ -26,22 +26,23 @@
...
   if (any(grepl("^\\{.*\\}$", tcode))) {
-    code <- sub("^\\{(.*)\\}$", "\\1", tcode)
+    tcode <- sub("^\\{(.*)\\}$", "\\1", tcode)
   }
 
+  parsed_code <- parse(text = tcode, keep.source = TRUE)
 
-  code <- parse(text = code, keep.source = TRUE)
-  pd <- utils::getParseData(code)
+  pd <- utils::getParseData(parsed_code)
```
gogonzo, line 45: Watch out @m7pr . Original `pd` has different rows order than one returned from `extract_calls(pd)`. `extract_calls` parsing `pd` recursively and returns a list of calls where rows are topologically ordered. Now we don't need `extract_calls` because the whole code is split into single codes and we just need to `parse` each of them. Problem is that if we just parse each code element we would get original `pd` which will break `extract_occurrence`. We either need to topologically sort pd or fix `extract_occurrence`. Quick example.
Please look at the difference below, assign operator is after assigned symbol in the original pd:

```
pd <- getParseData(parse(text ="x <- y", keep.source = TRUE))
teal.code:::extract_calls(pd)
#   line1 col1 line2 col2 id parent       token terminal text
# 7     1    1     1    6  7      0        expr    FALSE     
# 1     1    1     1    1  1      3      SYMBOL     TRUE    x
# 3     1    1     1    1  3      7        expr    FALSE     
# 2     1    3     1    4  2      7 LEFT_ASSIGN     TRUE   <-
# 4     1    6     1    6  4      6      SYMBOL     TRUE    y
# 6     1    6     1    6  6      7        expr    FALSE     
# > teal.code:::extract_calls(pd)
# [[1]]
#         token text id parent
# 7        expr       7      0
# 3        expr       3      7
# 2 LEFT_ASSIGN   <-  2      7
# 6        expr       6      7
# 1      SYMBOL    x  1      3
# 4      SYMBOL    y  4      6
```

m7pr: @gogonzo did you try running `extract_occurrence` on it?

```r
> teal.code:::extract_occurrence(teal.code:::extract_calls(pd))
[[1]]
[1] "x"  "<-" "y" 
```

gogonzo: We can discuss this once `get_code_dependency` works again. I put this comment based on the assumption that you're going to put `occurrence` and `side_effect` as a attribute of the `code` in `eval_code`. Let's ignore this comment at this moment, it might be relevant when you propose change of the `get_code_dependency`/`get_code`

R/qenv-extract.R
m7pr: 6338ce13a034595c19bac9c305cf607d00d65073

m7pr: 6338ce13a034595c19bac9c305cf607d00d65073

R/utils-get_code_dependency.R
```diff
@@ -108,25 +110,24 @@ find_call <- function(call_pd, text) {
...
       rbind(
-        pd[pd$id == parent, c("token", "text", "id", "parent")],
+        pd[pd$id == parent, ],
         get_children(pd = pd, parent = parent)
       )
     }
   )
   calls <- Filter(function(call) !(nrow(call) == 1 && call$token == "';'"), calls)
   calls <- Filter(Negate(is.null), calls)
-  calls <- fix_shifted_comments(calls)
```
m7pr, line 121: I think this might be needed.

This test is failing.

```r
testthat::test_that("get_code does not break if @linksto is put in the last line", {
  # In some cases R parses comment as a separate expression so the comment is not
  # directly associated with this line of code. This situation occurs when `eval` is in the last
  # line of the code. Other cases are not known but are highly probable.
  code <- c(
    "expr <- quote(x <- x + 1)",
    "x <- 0",
    "eval(expr) #@linksto x"
  )
  q <- eval_code(qenv(), code)
  testthat::expect_identical(
    get_code_g(q, names = "x"),
    code
  )
})
```

m7pr: On it!

R/utils-get_code_dependency.R
```diff
@@ -108,25 +110,24 @@ find_call <- function(call_pd, text) {
 #' @noRd
 extract_calls <- function(pd) {
   calls <- lapply(
-    pd[pd$parent == 0, "id"],
+    pd[pd$parent == 0 & pd$token != "COMMENT", "id"],
```
m7pr, line 113: I think it also removes comments with `# @linksto` tag, that span to the new line.

```r
tcode <- c("expr <- quote(x <- x + 1)", "x <- 0", "eval(expr) #@linksto x")
parsed_code <- parse(text = tcode, keep.source = TRUE)
pd <- utils::getParseData(parsed_code)
pd <- normalize_pd(pd)

```

image

R/utils-get_code_dependency.R
m7pr: Connected to https://github.com/insightsengineering/teal.code/pull/216/files#r1827636017

m7pr: Brought back in here 0afef8388f8804df5d6006078edd650bb8bee52a

R/utils-get_code_dependency.R
m7pr: Brought back in here 0afef8388f8804df5d6006078edd650bb8bee52a

tests/testthat/test-qenv_within.R
m7pr: 86e5d9588b732e4a4ee635c08db83dc645f73235

m7pr:
R CMD is failining, but will stop failining once this is merged https://github.com/insightsengineering/teal.code/pull/222

m7pr:
I am done with fixing the codebase after the latest updates to `split_code` functionality. The minor issue is that we introduced some new whitespaces somewhere. I think we can tackle this at the end. Not sure where to look up for it. Check out output for b, c, and the line after c.

Now I will try to incorporate `extract_occurence` and `side_effects` in `eval_code` as attributes of the `@code` slot.

``` r
library(teal.code)
code <- "
    # initial comment
    a <- 1
    b <- 2 # inline comment
    c <- 3
    # inbetween comment
    d <- 4
    # finishing comment
  "

q <- eval_code(qenv(), code)
cat(code)
#> 
#>     # initial comment
#>     a <- 1
#>     b <- 2 # inline comment
#>     c <- 3
#>     # inbetween comment
#>     d <- 4
#>     # finishing comment
#> 
cat(get_code(q))
#> 
#>     # initial comment
#>     a <- 1
#>      b <- 2 # inline comment
#>      c <- 3
#>      # inbetween comment
#>     d <- 4
#>     # finishing comment
#> 
```

<sup>Created on 2024-11-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

gogonzo:
> `cat(get_code(q))`

Because cat of a vector adds a space between elements. `get_code` used to return single character and it should still do it (otherwise you might even have more to fix in a downstram). 
It should rather be `get_code(q) |> paste(collapse = "")`

m7pr:
@gogonzo pheew, oh this is great. So we are good :) I raised another PR in here https://github.com/insightsengineering/teal.code/pull/223 where I moved `side_effects` and `occurrence` from `get_code_dependency` to `eval_code` but I don't think we removed a lot of code. I think we rather added code and complicated the print of `qenv@code`

tests/testthat/test-qenv_eval_code.R
```diff
@@ -88,22 +88,25 @@ testthat::test_that("an error when calling eval_code returns a qenv.error object
...
   testthat::expect_s4_class(q, "qenv")
   testthat::expect_equal(
-    q@warnings,
-    c("", paste(rep("> \"ff\" is not a graphical parameter\n", 4), collapse = ""))
+    lapply(q@code, attr, "warning"),
+    list(NULL, "> \"ff\" is not a graphical parameter\n")
   )
 })
 
 testthat::test_that("eval_code with a vector of code produces one warning element per code element", {
```
gogonzo, line 104: description no longer relevant. Warnings are the attribute of the code elements. It rather should be
`eval_code associates warnings with call by adding attribute to code element`

tests/testthat/test-qenv_extract.R
```diff
@@ -0,0 +1,68 @@
...
+    x <- 1
+    a <- 2
+  })
+  testthat::expect_warning(
+    testthat::expect_equal(data[c("y", "a")], data["a"]),
+    "Some elements of 'names' do not exist in 'qenv'. Skipping those: y."
+  )
+})
+
+testthat::test_that("`[.` extracts proper objects", {
```
gogonzo, line 24:
```suggestion
testthat::test_that("`[.` subsets environment and code to specified object names", {
```

tests/testthat/test-qenv_extract.R
```diff
@@ -0,0 +1,68 @@
...
+  q <- qenv()
+  code <- c("x<-1", "a<-1;b<-2")
+  q <- eval_code(q, code)
+  object_names <- c("x", "a")
+  qs <- q[object_names]
+  testthat::expect_true(all(ls(get_env(qs)) %in% object_names))
+})
+
+testthat::test_that("`[.` extract proper code", {
```
gogonzo, line 33: proper code is not expectation - please be more specific

tests/testthat/test-qenv_extract.R
```diff
@@ -0,0 +1,68 @@
...
+  q <- eval_code(q, code)
+  object_names <- c("x", "a")
+  qs <- q[object_names]
+  testthat::expect_identical(
+    unlist(qs@code),
+    c("x<-1\n", "a<-1;")
+  )
+})
+
+testthat::test_that("`[.` preservers comments in the code", {
```
gogonzo, line 45: There should be multiple comment-preserving-tests

- Comments are associated with the call from the same line (within same eval_code)
- Comments are associated with following call (within same eval_code)
- Comments in the end of the src are associated with the last call (within same eval_code)
- Comments alone (without code) are added as a separate code element

```suggestion
testthat::test_that("`[.` comments are preserved in the code and associated with the following call", {
```

tests/testthat/test-qenv_eval_code.R
```diff
@@ -41,21 +41,21 @@ testthat::test_that("getting object from the package namespace works even if lib
 testthat::test_that("eval_code works with character", {
   q1 <- eval_code(qenv(), "a <- 1")
 
-  testthat::expect_identical(q1@code, "a <- 1")
```
gogonzo, line 44: "Nature" of the `eval_code` and `@code` has changed and tests needs to be rewritten to fit better a current state. `@code` holds call asis, which means that many tests can be written like so:

```
# testthat {
  code <- "some code"
  q <- eval_code(<qenv>, code)
 # testthat::expect_identical(<drop_attributes>(q@code), as.list(code)) 
 # testthat::expect_identical(get_code(code), code)
}

```

R/qenv-eval_code.R
```diff
@@ -28,24 +28,23 @@
 setGeneric("eval_code", function(object, code) standardGeneric("eval_code"))
 
 setMethod("eval_code", signature = c("qenv", "character"), function(object, code) {
-  id <- sample.int(.Machine$integer.max, size = 1)
-
-  object@id <- c(object@id, id)
+  parsed_code <- parse(text = code, keep.source = TRUE)
```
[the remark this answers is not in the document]
m7pr: It is used below in 

```r
parsed_code <- parse(text = code, keep.source = TRUE)
if (length(parsed_code) == 0) {
  return(object)
}
```

and prevents cases when code is `""` or `character(0)`.

We could change this whole condition to

```r
if (identical(code, "") || length(code) == 0) {
  return(object)
}
```

but I'm worried there are some weird cases we don't cover but parser covers when parsing.

tests/testthat/test-qenv_extract.R
m7pr: 084382a3b6021705c6e1ab968ae4fcc26074d426

m7pr:
Hey @gogonzo tests needs to be rewritten - I will ask for a direct review. For now I am marking this `Ready for the review` so that tests can run for this branch. But I will let you know when I'm done with adjusting tests

R/qenv-get_warnings.R
```diff
@@ -33,7 +33,9 @@ setGeneric("get_warnings", function(object) {
 })
 
 setMethod("get_warnings", signature = c("qenv"), function(object) {
-  if (all(object@warnings == "")) {
+  warnings <- lapply(object@code, "attr", "warning")
```
gogonzo, line 36: Too many as.list/unlist. Please optimise
```suggestion
  warnings <- lapply(object@code, "attr", "warning")
  idx_warn <- which(sapply(warnings, Negate(is.null)))
```

tests/testthat/test-qenv_get_code.R
```diff
@@ -1,11 +1,15 @@
+get_code_g <- function(qenv, ...) {
+  gsub("\n", "", get_code(qenv, ...), fixed = TRUE)
+}
+
 testthat::test_that("get_code returns code (character by default) of qenv object", {
   q <- qenv() |>
     eval_code(quote(x <- 1)) |>
     eval_code(quote(y <- x))
-  testthat::expect_equal(get_code(q), paste(c("x <- 1", "y <- x"), collapse = "\n"))
+  testthat::expect_equal(get_code_g(q), c("x <- 1", "y <- x"))
```
gogonzo, line 9: This is a breaking change. `get_code` used to return concatenated `character`.

gogonzo (changes requested):
Breaking change in `get_code` - it should return concatenated character and returns a vector now.

R/qenv-get_warnings.R
m7pr: Thanks, changed in here https://github.com/insightsengineering/teal.code/pull/216/commits/7bd1c2d376e76427dab232585c06938d8e266095

tests/testthat/test-qenv_get_code.R
m7pr: I think I did so many back and forths on this PR with so many features, that I just lost track. Thanks for catching
I reverted the output of `get_code` so it returns concatenated text in the same form as in input code https://github.com/insightsengineering/teal.code/pull/216/commits/7fa236931185c91dccc7df526696732991280625

tests/testthat/test-qenv_eval_code.R
m7pr: Now tests are written in this format

```r
testthat::expect_identical(get_code(code), paste(code, collapse = "\n"))
```

The

```
testthat::expect_identical(<drop_attributes>(q@code), as.list(code)) 
```

part from the suggestion is not true, as `code` can have multiple calls in one element and can have length 1, where `q@code` will have length greater than one for input with multiple calls.

tests/testthat/test-qenv_extract.R
m7pr: Hey, will continue on this one now

m7pr:
@gogonzo I need you to copy and paste and send this in a new comment

> I have read the CLA Document and I hereby sign the CLA

tests/testthat/test-qenv_eval_code.R
```diff
@@ -88,41 +88,85 @@ testthat::test_that("an error when calling eval_code returns a qenv.error object
     unname(q$trace),
     c("x <- 1", "y <- 2", "z <- w * x")
   )
-  testthat::expect_equal(q$message, "object 'w' not found \n when evaluating qenv code:\nz <- w * x")
+  testthat::expect_equal(q$message, "object 'w' not found \n when evaluating qenv code:\nexpression(z <- w * x)")
```
m7pr, line 91: This is failining and now contains a bit more.
It is becuase in `eval_code` when we calculate the message error we use
`current_call` that has `keep.source = TRUE` in the `parse`.

Maybe we can substitute `current_call` with `current_code`

```r
x[1]: "object 'w' not found \n when evaluating qenv code:\nstructure(expression(z 
x[1]: <- w * x), srcfile = <environment>, wholeSrcref = structure(c(1L, 0L, 2L, 0L
x[1]: , 0L, 0L, 1L, 2L), srcfile = <environment>, class = \"srcref\"))"
y[1]: "object 'w' not found \n when evaluating qenv code:\nexpression(z <- w * x)"
```

m7pr: Fixed in here 20b2c8fc00128c37d46d4eeeed403e72a658a4fc

m7pr: and updated in here b4e119c98bbe1555aec552920a4ed72c3027be3c

gogonzo:

tests/testthat/test-qenv_extract.R
m7pr: Hey @gogonzo we had few tests for comments, but in other file. I moved them and extended them. They now leave in eval_code e28b937bdc3e927c576de4d86eb163d26dbf17ed

m7pr:
Hey @gogonzo I provided fixes for the comments/issues that you found. I hope this is now in a better state.

Frankly I need to admit that this becomes a bit too big for me right now and I better get this closed and start new small issues if we need more improvements. We changed a lot of big parts of the code. This scope is too much to handle in one issue and is too much to have an easy review. 

It also takes me a more time than usual to restart the context to this branch, when I get back after reviewing other PRs.

tests/testthat/test-qenv_get_code.R
```diff
@@ -632,14 +627,15 @@ testthat::test_that("detects cooccurrence properly even if all objects are on lh
...
-    paste(code, collapse = "\n")
+    pasten(code)
   )
 })
 
 
 # @ ---------------------------------------------------------------------------------------------------------------
 
 testthat::test_that("understands @ usage and do not treat rhs of @ as objects (only lhs)", {
+  testthat::skip("This breaks on CI but not locally with: Error: cannot add bindings to a locked environment")
```
m7pr, line 638: Previously this was tested just on `code` that was a character and there was no evaluation.
Now we use `eval_code` and the code gets evaluated.

m7pr: Fixed in here 7cf13ac0b003be831c3d7dc0d9027dfecc99ba7a

R/qenv-eval_code.R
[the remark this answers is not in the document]
gogonzo: What are the cases here:
```r
c1 <- parse(text = "\n\n ", keep.source = TRUE) 
c1
# expression()
length(c1)
# [1] 0

c2 <- parse(text = "# only a comment", keep.source = TRUE)
c2
# expression()
length(c2) 
# [1] 0



split_code("\n\n") # error
split_code("# only a comment") # error
split_code("a <- 1")
```

The escape condition is correct. You can create an issue for `eval_code` to not handle the comments or fix split_code to accept empty calls.

R/qenv-get_code.R
```diff
@@ -124,13 +122,9 @@ setMethod("get_code", signature = "qenv", function(object, deparse = TRUE, names
   }
 
   if (deparse) {
-    if (length(code) == 0) {
-      code
-    } else {
-      paste(code, collapse = "\n")
-    }
+    gsub(";\n", ";", paste(gsub("\n$", "", unlist(code)), collapse = "\n"))
```
gogonzo, line 125: - remove "\n" to `paste(collapse = "\n")` is redundant or even wrong (it will remove "\n" from the last item). 
- also is `";\n"` a problem? It is a valid call


```suggestion
    paste(unlist(code), collapse = "")
```

R/qenv-get_warnings.R
```diff
@@ -33,7 +33,11 @@ setGeneric("get_warnings", function(object) {
...
+  idx_warn <- which(sapply(warnings, Negate(is.null)))
+  warnings <- warnings[idx_warn]
+  code <- object@code[idx_warn]
+  if (length(unlist(warnings)) == 0) {
```
gogonzo, line 40:
```suggestion
  if (length(warnings) == 0) {
```

R/qenv-get_code.R
gogonzo: Ok, I'm wrong here. There is possibility that somebody provides a single call without terminator. The code will be preserved as is, for example `"a <- 1"`. Ignore my comment here ⬆️

m7pr: I experimented with this a bit before I pushed it

That's what we get with `paste(unlist(code), collapse = "")`

```r
q <- qenv()
q <- eval_code(q, quote(x <- 1))
q <- eval_code(q, quote(y <- x))
get_code(q)
[1] "x <- 1y <- x"
```

When we use `collapse = "\n"`, then this is fine

```
"x <- 1ny <- x"
```

but 

```r
q <- qenv()
q <- eval_code(q, quote(x <- 1))
q <- eval_code(
  q,
  quote({
    y <- x
    z <- 5
  })
)
```

gives 2 `\n`s after x

```r
> get_code(q)
[1] "x <- 1\ny <- x\n\nz <- 5"
```

```
> cat()
x <- 1
y <- x

z <- 5
```


Lastly, the `";\n", ";"` substitution is for cases like

```r
testthat::test_that("get_code returns code with comments and empty spaces", {
  code <- "
    # header comment after white space

    a <- 1L; b <- 2 #inline comment


    c <- 3
    # closing comment
    "
  q <- eval_code(qenv(), code)
  testthat::expect_equal(get_code(q), code)
})
```

image


Those are first few tests in `test-qenv_get_code`.

We do actually a manual imputation for cases with `:` in the middle line. We add `\n`. that's why we need to deal with `;` and `\n` when extracting the code. Those are those lines https://github.com/insightsengineering/teal.code/pull/216/files#diff-66a61facc4f5ca86215d63af40e2645b364950439051784916316ef78d5073c8R476-R479

> it will remove "\n" from the last item

I think we can leave with a situation where the last one is used, but should not be. This is the case where it was the last one in the code. All other cases are tested

tests/testthat/test-qenv_get_code.R
```diff
@@ -45,15 +47,28 @@ testthat::test_that("get_code called with qenv.error returns error with trace in
...
+
+    a <- 1L; b <- 2 #inline comment
+
+
+    c <- 3
+    # closing comment
+    "
+  q <- eval_code(qenv(), code)
+  testthat::expect_equal(get_code(q), code)
+})
```
gogonzo, line 62: This test makes me happy 💪

tests/testthat/test-qenv_get_code.R
```diff
@@ -649,26 +644,29 @@ testthat::test_that("understands @ usage and do not treat rhs of @ as objects (o
     "a@x <- x@a"
   )
   q <- qenv()
-  q@code <- code # we don't use eval_code so the code is not run
+  code_split <- as.list(split_code(paste(code, collapse = "\n")))
```
gogonzo, line 647: we don't test private utils. If it doesn't work with `eval_code` then we should test for error

m7pr: I am testing the code_parser/graph_parser part in here as I think it's pretty important to test how it understands `@`. 
If you don't eval the code, and only analyze the code as a character this test make sense. For `teal_data` it was possible to create `teal_data` with the code but without the code execution. For `qenv` this is not possible. I would like to see how parser handles `@` in tests

tests/testthat/test-qenv_eval_code.R
```diff
@@ -91,38 +91,178 @@ testthat::test_that("an error when calling eval_code returns a qenv.error object
...
+      )
+    )
+  }
+)
+
+
+
+# comments --------------------------------------------------------------------------------------------------------
+
+
```
m7pr, line 178:
```suggestion
```

tests/testthat/test-qenv_eval_code.R
```diff
@@ -91,38 +91,175 @@ testthat::test_that("an error when calling eval_code returns a qenv.error object
...
+    paste(code, collapse = "\n")
   )
 })
 
-testthat::test_that("eval_code returns a qenv object with empty messages and warnings when none are returned", {
-  q <- eval_code(qenv(), quote("iris_data <- head(iris)"))
-  testthat::expect_s4_class(q, "qenv")
-  testthat::expect_equal(q@messages, "")
-  testthat::expect_equal(q@warnings, "")
+testthat::test_that("comments alone passed to eval_code are skipped", {
```
m7pr, line 258: @gogonzo this is pretty important - comments passed alone are skipped. Unsure if we want to handle this like that

tests/testthat/test-qenv_get_code.R
gogonzo: If it is the case for `teal_data` then the test shouldn't be here ;]

gogonzo: Please ignore my comment then. It is fine, I'll highlight this another time when somebody would touch this part ;]

m7pr: coolio

R/qenv-eval_code.R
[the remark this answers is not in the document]
m7pr: I missed this comment. I provided a fix in eval_code to handle comments

3ba65f2727f6247d69e26a4eced04d167e994c4a

R/qenv-eval_code.R
```diff
@@ -96,8 +96,12 @@ setMethod("eval_code", signature = c("qenv", "language"), function(object, code)
...
 
 setMethod("eval_code", signature = c("qenv", "expression"), function(object, code) {
-  # todo: if has srcfile then get original text!
-  eval_code(object, code = paste(lang2calls(code), collapse = "\n"))
+  srcref <- attr(code, "wholeSrcref")
+  if (length(srcref)) {
+    eval_code(object, code = paste(attr(code, "wholeSrcref"), collapse = "\n"))
+  } else {
+    eval_code(object, code = paste(lang2calls(code), collapse = "\n"))
+  }
```
m7pr, line 104: nice

m7pr:
Thanks @gogonzo for the updates!
I see `get_messages` and `get_warnigs` have a lot of common, they just extract different name of the attribute and they contain different header in the output text.

Do you think those could share the same core and be wrappers to one function that takes the name of the attribute as an input?

gogonzo:
> Do you think those could share the same core and be wrappers to one function that takes the name of the attribute as an input? 

good point 👍  I'll implement and fix CICD in the next commit

R/qenv-extract.R
```diff
@@ -0,0 +1,50 @@
...
+#' # Subsetting
+#' q <- qenv()
+#' q <- eval_code(q, "a <- 1;b<-2")
+#' q["a"]
+#' q[c("a", "b")]
+#'
+#' @rdname qenv
+#'
+#' @export
+`[.qenv` <- function(x, names, ...) {
```
averissimo, line 19: An `NA` artifact is created when subsetting the module's datanames with dot prefix such as `.raw_data`.

```r
pkgload::load_all("teal.data")
pkgload::load_all(".")

app <- init(
  data = teal_data() |> within(iris <- iris),
  modules = example_module()
) |> shiny::runApp()
```

image

image

R/qenv-extract.R
```diff
@@ -0,0 +1,50 @@
...
+  checkmate::assert_class(names, "character")
+  possible_names <- ls(get_env(x), all.names = TRUE)
+  names_warn <- setdiff(names, possible_names)
+  names <- intersect(names, possible_names)
+  if (!length(names)) {
```
gogonzo, line 24: what if?

```r
q <- within(qenv(), {
	a <- iris
	b <- a
	plot(a)
	rm(a)
}) |> dev_suppress()

q["a"]


```

m7pr: Sure, happy to chat whenever you can! Ready when you are

gogonzo (approved):
🔥

Issue #164, opened 2023-11-13
[Feature Request]: New `$` method to access objs from env


pawelru:
### Feature description

Functionality identical to the `[[` but `$` would be a little bit more concise: `x["foo"]` vs `x$foo`.

todo:
- new method
- docs
- tests
- decide which one to promote as a _main_ one: `[[` or `$` - if the latter then change readme / examples / docs / vignettes and possibly docs in other packages as well

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

m7pr:
@pawelru maybe we can have both?

And then, is it just

```r
`$.qenv` <- function(object, name) object[[name]]
```

pawelru:
Yeah - we can definitely have both. This is a request for extension - not necessarily a replacement. However, it would be good to have one _preffered_ way of doing this so that our set of examples could follow a similar logic and be simpler to the end users

llrs-roche:
With `$` working one could use .DollarNames.qenv to enable autocomplete (See ?.DollarNames with the dot) and recommend that way to access the data. This is relatively easy to do given all the other methods implemented but could have a big impact on how users write with the class.

PR #218, opened 2024-10-28
Change `qenv` as `environment` "type" -- adds `names(qenv/qenv.error)`, `get()`  and `$` S3 methods

averissimo:
# Pull Request

- Part of https://github.com/insightsengineering/teal.data/issues/333
- Fixes https://github.com/insightsengineering/teal.code/issues/221 
- closes https://github.com/insightsengineering/teal.code/issues/164
- Companion of https://github.com/insightsengineering/teal.data/pull/347

#### Changes description

- `qenv` S4 class inherits from `environment` data class
- Removes `@env` slot in favor of `qenv`
- Replace all instances of `@env` with `@.xData` (slot created by parent class)
- All functions/methods that work for `environment` class are supported natively in `qenv`


R/qenv-names.R
```diff
@@ -0,0 +1,33 @@
...
+#'   CO2 <- CO2
+#' })
+#' names(q2)
+#'
+#' @export
+names.qenv <- function(x) {
+  ls(get_env(x))
+}
+
+#' @rdname names.qenv
```
gogonzo, line 29: We need to decide where to put `names.qenv`. At this moment `within`, `eval_code`, `get_code` are part of `qenv` documentation and they are described in the sections. Problem with some methods (`names`, `within`) is that they have generic in base and it is not obvious that one needs to call `names.qenv`.

gogonzo (commented):
Please test cover new functions

NAMESPACE
```diff
@@ -1,6 +1,9 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method("$",qenv.error)
+S3method("$<-",qenv.error)
 S3method("[[",qenv.error)
+S3method("[[<-",qenv.error)
```
gogonzo, line 6: Why is it needed?

averissimo: I'd like to preface that I'm more than happy to remove as this is \*\*totally\*\* nitpicking.

My reasoning adding this would be to make the API consistent

The 2 snippets below shows the \*\***released/main**\*\* behavior and the assignments (`<-`) seem inconsistent with `qenv`. 

It seems odd that with `qenv.error` the script runs a few more commands.

### `qenv` object without errors

``` r
library(teal.code)
q <- qenv() |> within(a <- 2)
q$something <- 2
#> Error in `$<-`(`*tmp*`, something, value = 2): no method for assigning subsets of this S4 class
q[["another"]] <- 3
#> Error in `[[<-`(`*tmp*`, "another", value = 3): [[<- defined for objects of type "S4" only for subclasses of environment

q$something
#> Error in q$something: $ operator not defined for this S4 class
q[["another"]]
#> object 'another' not found
#> NULL
```

<sup>Created on 2024-10-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### `qenv` object when there's an error in `within` code

``` r
library(teal.code)
q <- qenv() |> within(stop())
q$something <- 2
q[["another"]] <- 3

q$something
#> [1] 2
q[["another"]]
#> Error:  
#>  when evaluating qenv code:
#> stop()
```

<sup>Created on 2024-10-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/qenv-eval_code.R
```diff
@@ -31,7 +31,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
   id <- sample.int(.Machine$integer.max, size = 1)
 
   object@id <- c(object@id, id)
-  object@env <- rlang::env_clone(object@env, parent = parent.env(.GlobalEnv))
+  object@.xData <- rlang::env_clone(object@.xData, parent = parent.env(.GlobalEnv))
```
averissimo, line 34: {rlang} functions don't play nicely with this approach, so we need to access the slot directly

``` r
#> ℹ Loading teal.code

q1 <- qenv()
rlang::is_environment(q1)
#> [1] FALSE
base::is.environment(q1)
#> [1] TRUE

rlang::env_clone(q1)
#> Error in `rlang::env_clone()`:
#> ! `env` must be an environment, not a <qenv> object.

rlang::is_environment
#> function(x) {
#>   typeof(x) == "environment"
#> }
#> <bytecode: 0x65127a0d6b08>
#> <environment: namespace:rlang>
typeof(q1)
#> [1] "S4"
```

<sup>Created on 2024-10-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/qenv-get_var.R
```diff
@@ -55,3 +55,29 @@ setMethod("[[", signature = c("qenv", "ANY"), function(x, i) {
     class = c("validation", "try-error", "simpleError")
   ))
 }
+
+#' @export
+`$.qenv.error` <- function(x, name) {
```
averissimo, line 60: Should we add this method and have it behave just like `[[`?

We need to allow `$` to retrieve existing properties of `qenv.error`, as `conditionMessage(x)` will try to access `x$message`, `x$call` and `x$trace`.

R/qenv-get_var.R
```diff
@@ -55,3 +55,29 @@ setMethod("[[", signature = c("qenv", "ANY"), function(x, i) {
...
+  # Must allow access of elements in qenv.error object (message, call, trace, ...)
+  # Otherwise, it will enter an infinite recursion with the `conditionMessage(x)` call.
+  if (name %in% names(x)) {
+    return(NextMethod("$", x))
+  }
+
+  class(x) <- setdiff(class(x), "qenv.error")
```
averissimo, line 67: Safeguard to prevent infinite recursion, this can be "safely" removed if needed

tests/testthat/test-qenv_constructor.R
```diff
@@ -1,19 +1,67 @@
+testthat::describe("qenv inherits from environment: ", {
```
averissimo, line 1: Added test that addressed your last comment before the "change"

averissimo (commented):
Some topics to discuss on this implementation

NAMESPACE
gogonzo: We don't need to fence all possible methods. There is large set of the methods inherited from `environment`, it is important that officially supported functions are mentioned in the docs of `qenv`. We need to mention inherited working methods but not add them. What I would add though to `qenv.error` is `$.qenv.error`,  `[[.qenv.error`, `as.list.qenv.error`, `ls.qenv.error`

averissimo: `ls.qenv.error` cannot be added, but it will still give an error, so no harm there

R/qenv-constructor.R
```diff
@@ -10,9 +10,31 @@
...
+#'
 #' `new_qenv()` (`r badge("deprecated")` and not recommended)
 #' can instantiate a `qenv` object with data in the environment and code registered.
 #'
+#' @section Environment:
+#'
+#' The `qenv` object behaves as an environment that is locked and can be used as
+#' as a argument to with many functions that accept `environment`, among the
+#' most relevant are: `names()`, `ls()`, `get()`, `exists()`, `parent.env()`,
+#' `{l,s,v}apply` family, `local`, `as.environment()`, `is.environment()`, `as.list()`, ...
```
gogonzo, line 22:
```suggestion
#' The `qenv` object behaves like an environment that is locked and one can use
#' some of the `base` functions dedicated to the `environment`. List of supported 
#' functions includes: 
#' `names()`, `ls()`, `get()`, `exists()`, `parent.env()`, `lapply`, `sapply`
#' `vapply`, `local`, `as.environment()`, `is.environment()`, `as.list()`, ...
#' We don't recommend using any function outside of the `teal.code` exports and these 
#' mentioned above.
```

R/qenv-constructor.R
```diff
@@ -10,9 +10,31 @@
...
+#' problematic function.
+#'
+#' Similarly, `rlang` functions related to environments cannot be used directly
+#' with `qenv` and should be used with `teal.code::get_env()`/`as.environment()`.
+#'
+#' ```r
+#' q <- qenv()
+#' rlang::env_clone(as.environment(q))
+#' rlang::env_clone(get_env(q))
+#' ```
```
gogonzo, line 36: I think "_We don't recommend using any function outside of the `teal.code` exports and these mentioned above_" is enough. Explaining that something doesn't work is adding more content to the simple statement. Also I don't see the point of guiding users what to do when `rlang` fails.

```suggestion
```

gogonzo (commented):
There are some `@env` remaining in `R/` directory . Also there shouldn't be any mention about internal slots (`regex:#'.+@(env|.xData|code|id|warnings|messages)` in the documentation `@` - please replace them with something implicit. 
(you can leave join docs unchanged - this is a bigger problem)

NAMESPACE
gogonzo: hmm, there is a difference between throwing error and returning error. `ls(<qenv.error>)` will throw while `names.qenv.error(<qenv.error>)` will return error. The first one needs to be tryCatched while second one not. We need to check if `ls(data())` in `teal` internals doesn't throw because of this (probably not as `teal` is protected against any types of errors)

R/qenv-constructor.R
```diff
@@ -23,9 +45,7 @@
 #'
```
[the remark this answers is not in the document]
averissimo: See [0a644988135a098665-af7054b837b](https://github.com/insightsengineering/teal.code/pull/218/files/0a644988135a09866546dc022cbab0a4f1266332..af7054b837b45be9271b667e2c8b9db0a769f285).

I may have gone a bit overboard by including these methods in the usage section, WDYT?

R/qenv-get_var.R
```diff
@@ -1,5 +1,9 @@
 #' Get object from `qenv`
 #'
+#' @description
+#' `r lifecycle::badge("superseded")` by native \R operators/functions:
```
averissimo, line 4: Should we mark this function and `join` as superseded?

gogonzo: yes 👍

R/qenv-constructor.R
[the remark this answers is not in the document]
gogonzo: Yup. I think it is okey. I would like to revisit this docs soon and rearrange the content (not in this PR)

NEWS.md
```diff
@@ -4,6 +4,7 @@
 
 * `get_code()` was extended with `names` parameter and allows the code extraction to be limited to objects stored in
 `qenv` but limited to `names`.
+* `qenv` inherits from the `environment` class, allowing to use `ls()`, `names()`, `as.environment()` and other functions on `qenv` objects.
```
gogonzo, line 7:
```suggestion
* `qenv` inherits from the `environment` class, allowing to use `ls()`, `names()`, `as.environment()` and other functions on `qenv` objects.
* `join()` method is deprecated, please use `c()` instead
* `get_var()` method is deprecated, please use `get`, `[[` or `$` instead.
```

R/qenv-get_var.R
gogonzo, line 4:
```suggestion
#' `r lifecycle::badge("deprecated")` by native \R operators/functions:
```

R/qenv-get_var.R
```diff
@@ -28,7 +31,7 @@ setGeneric("get_var", function(object, var) {
 
 setMethod("get_var", signature = c("qenv", "character"), function(object, var) {
   tryCatch(
-    get(var, envir = object@env, inherits = FALSE),
+    get(var, envir = object@.xData, inherits = FALSE),
```
gogonzo, line 34:
```suggestion
   lifecycle::deprecate_soft("0.5.1", "get_var()", "get()")
    get(var, envir = object@.xData, inherits = FALSE),
```

R/qenv-join.R
```diff
@@ -74,8 +78,8 @@
 #'    # Error message will occur
 #'
 #'    # Check the value of temporary variable i in both objects
-#'    x@env$i # Output: 2
-#'    y@env$i # Output: 3
+#'    x@.xData$i # Output: 2
+#'    y@.xData$i # Output: 3
```
gogonzo, line 82:
```suggestion
#'    x$i # Output: 2
#'    y$i # Output: 3
```

gogonzo (commented):
Please update the vignette. There are calls which should be replaced:
`ls(get_env(my_qenv))` -> `ls(my_qenv)`
`join(join_q)` -> `c(join_q)`
`q_message@messages` -> `get_messages(q_message)`
`q_warning@warnings` -> `get_warnings(q_warning)`

tests/testthat/test-qenv_constructor.R
```diff
@@ -1,19 +1,67 @@
...
+    testthat::expect_identical(parent.env(q), parent.env(.GlobalEnv))
+  })
 })
 
-testthat::test_that("parent of qenv environment is locked", {
+testthat::test_that("qenv environment is locked", {
   q <- qenv()
-  testthat::expect_error(q@env$x <- 1, "cannot add bindings to a locked environment")
+  testthat::expect_error(q@.xData$x <- 1, "cannot add bindings to a locked environment")
 })
```
gogonzo, line 67: duplicated test with the one from line 31 - except that here you're trying to modify slot directly and there using environment inheritance. I think checking both scenarios are not necessary
```suggestion
```

tests/testthat/test-qenv_eval_code.R
```diff
@@ -15,14 +15,14 @@ testthat::test_that("eval_code doesn't have access to environment where it's cal
...
-  testthat::expect_identical(parent.env(q1@env), parent.env(.GlobalEnv))
+  testthat::expect_identical(parent.env(q1@.xData), parent.env(.GlobalEnv))
 
   q2 <- eval_code(q1, quote(a <- 1L))
-  testthat::expect_identical(parent.env(q2@env), parent.env(.GlobalEnv))
+  testthat::expect_identical(parent.env(q2@.xData), parent.env(.GlobalEnv))
   q3 <- eval_code(q2, quote(b <- 2L))
-  testthat::expect_identical(parent.env(q3@env), parent.env(.GlobalEnv))
+  testthat::expect_identical(parent.env(q3@.xData), parent.env(.GlobalEnv))
 })
```
gogonzo, line 26: duplicated test - already introduced in line 52 
```suggestion
```

tests/testthat/test-qenv_eval_code.R
```diff
@@ -72,11 +72,15 @@ testthat::test_that("eval_code works with quoted code block", {
...
+  testthat::expect_equal(q1@.xData, list2env(list(a = 1, b = 2)))
 })
 
 testthat::test_that("eval_code fails with unquoted expression", {
-  testthat::expect_error(eval_code(qenv(), a <- b), "object 'b' not found")
+  b <- 3
+  testthat::expect_error(
+    eval_code(qenv(), a <- b),
+    "unable to find an inherited method for function .eval_code. for signature"
+  )
```
gogonzo, line 83: I don't understand what have changed that you have added `b <- 3`. In general I think test is not needed at all. `eval_code` supports certain types of inputs, providing failing call `a <- b` doesn't make sense - we don't need to test R execution precedence

tests/testthat/test-qenv_join.R
```diff
@@ -60,15 +60,15 @@ testthat::test_that("join does not duplicate code but adds only extra code", {
...
 
   testthat::expect_identical(
     q@code,
     c("iris1 <- iris\nmtcars1 <- mtcars", "iris2 <- iris", "mtcars2 <- mtcars")
   )
 
   testthat::expect_equal(
-    as.list(q@env),
+    as.list(q@.xData),
     list(iris1 = iris, iris2 = iris, mtcars1 = mtcars, mtcars2 = mtcars)
```
gogonzo, line 72: I'd rather check this way - pure result vs expectation
```suggestion
    q@.xData,
    list2env(list(iris1 = iris, iris2 = iris, mtcars1 = mtcars, mtcars2 = mtcars))
```

tests/testthat/test-qenv_constructor.R
averissimo: 620849bd

tests/testthat/test-qenv_eval_code.R
averissimo: 1e256810

R/qenv-get_var.R
gogonzo: I need to think more about lines 66-70

R/qenv-join.R
```diff
@@ -74,10 +78,10 @@
 #'    # Error message will occur
 #'
 #'    # Check the value of temporary variable i in both objects
-#'    x@env$i # Output: 2
-#'    y@env$i # Output: 3
+#'    x$i # Output: 2
+#'    y$i # Output: 3
 #'    ```
-#'    `join()` fails to provide a proper result because of the temporary variable `i` exists
+#'    `c()` and `join()` fails to provide a proper result because of the temporary variable `i` exists
```
gogonzo, line 84:
```suggestion
#'    `c()` fails to provide a proper result because of the temporary variable `i` exists
```

R/qenv-names.R
averissimo: If we want to keep them visible, they need to be attached to `qenv` or other docs.

Otherwise, they have to be accessed indirectly (`see also` sections or via manual search)

tests/testthat/test-qenv_eval_code.R
gogonzo: Yes, let's delete 👍

R/qenv-get_var.R
```diff
@@ -44,14 +48,28 @@ setMethod("get_var", signature = c("qenv.error", "ANY"), function(object, var) {
...
+`$.qenv.error` <- function(x, name) {
+  # Must allow access of elements in qenv.error object (message, call, trace, ...)
+  # Otherwise, it will enter an infinite recursion with the `conditionMessage(x)` call.
+  if (exists(name, x)) {
+    return(NextMethod("$", x))
+  }
+
+  class(x) <- setdiff(class(x), "qenv.error")
+  stop(errorCondition(
+    list(message = conditionMessage(x)),
```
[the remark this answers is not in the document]
averissimo: We need to keep some of the classes otherwise, `conditionMessage` fails

```r
Browse[1]> conditionMessage(unclass(x))
#> Error in UseMethod("conditionMessage") : 
#>   no applicable method for 'conditionMessage' applied to an object of class "list"
```

R/qenv-class.R
```diff
@@ -14,11 +14,60 @@
...
+    lockEnvironment(new_xdata, bindings = TRUE)
+
+    # .xData needs to be unnamed as the `.environment` constructor allows at
+    # most 1 unnamed formal argument of class `environment`.
+    # See methods::findMethods("initialize")$.environment
+    .Object <- methods::callNextMethod( # nolint: object_name.
+      # Mandatory use of `xData` to build a correct .Object@.xData
+      .Object, new_xdata,
+      code = code, messages = messages, warnings = warnings, id = id, ...
+    )
```
averissimo, line 67: This call is essential with `.xData` argument being mandatory!

Without it, the `@.xData` is the same between different instances on `qenv` (scary stuff and why I was delayed today to push this)

``` r
setClass(
  "an_env",
  contains = "environment"
)

setMethod(
  "initialize",
  signature = "an_env",
  definition = function(.Object, ...) {
    .Object
  }
)
q <- methods::new("an_env")
q@.xData$a <- 2
as.list(q)
#> $a
#> [1] 2

q2 <- methods::new("an_env")
as.list(q2)
#> $a
#> [1] 2
```

<sup>Created on 2024-11-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

gogonzo: This is interesting finding. I wonder how it works exactly as environment is not defined anywhere by us in the package. It is created inside of the function. Let's have a quick call after daily if you can

averissimo: Let's!

gogonzo: Okey, I see now how it works.

tests/testthat/test-qenv-class.R
```diff
@@ -0,0 +1,24 @@
...
+    expect_error(methods::new("qenv", id = 1L), "@code and @id slots must have the same length\\.")
+  })
+
+  testthat::it("throws error when .xData is not an environment", {
+    expect_error(methods::new("qenv", .xData = 2), "Must be an environment, not 'double'\\.")
+  })
+
+  testthat::it("throws error when code is not language or character object", {
+    expect_error(methods::new("qenv", code = 2), "`code` must be a character or language object\\.")
+  })
```
gogonzo, line 23:
```suggestion
  })
  testthat::it("initialized qenv(s) have different environments", {
    testthat::expect_false(identical(qenv()@.xData, qenv()@.xData)) 
  })
```

averissimo: Awesome!

R/qenv-eval_code.R
gogonzo: Not a blocker for this approach 👍

R/qenv-get_var.R
gogonzo: Thanks for figuring this out 👍

README.md
```diff
@@ -62,9 +62,9 @@ my_qenv
 #> Parent: <environment: package:teal.code>
 #> Bindings:
 #> • x: <dbl> [L]
-get_env(my_qenv)
+as.environment(my_qenv)
 #> <environment: 0x00000225cc85c7a0>
-ls(get_env(my_qenv))
+names(my_qenv))
```
gogonzo, line 67:
```suggestion
names(my_qenv)
```

R/qenv-get_env.R
```diff
@@ -13,7 +13,8 @@
 #'   b <- data.frame(x = 1:10)
 #' })
 #' get_env(q1)
-#' ls(get_env(q1))
+#'
+#' names(get_env(q1)) # list objects in the environment
```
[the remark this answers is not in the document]
averissimo: Given this is for the `get_env()` documentation, I think it should stay as is, or:

```suggestion
#' names(get_env(q1)) # list objects in the environment which is equivalent to names(q1)
```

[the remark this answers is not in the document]
gogonzo: feel free to commit. I don't like the idea of presenting every possible way. I'd stick only to `names(<qenv>)`. One should know that when extracting `environment` then the environment methods can be applied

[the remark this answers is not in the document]
averissimo: I like that, it's better to guide towards using `<qenv>` as an environment and not as container for an environment. Commiting your suggestion.

R/qenv-get_env.R
```diff
@@ -14,7 +14,7 @@
 #' })
 #' get_env(q1)
 #'
-#' names(get_env(q1)) # list objects in the environment
+#' names(q1) # list objects
```
averissimo, line 17:
```suggestion
```

R/qenv-get_env.R
```diff
@@ -13,7 +13,8 @@
 #'   b <- data.frame(x = 1:10)
 #' })
 #' get_env(q1)
-#' ls(get_env(q1))
+#'
+#' names(q1) # list objects
 #'
```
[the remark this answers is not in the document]
gogonzo: yup ;]

PR #223, opened 2024-11-04
assign `side_effects` and `occurrence` as attributes of `@code`

m7pr:
Part of #216 

Changes:
- [x] moved `dependency` extraction from `get_code_dependency` to `eval_code`
- [x] removed `extract_code_graph`
- [x] extended documentation of `qenv` with 1 new attributes: `dependency`, `occurrence`
- [x] merged `side_effects` and `occurrence` inside `eval_code` as they were previously joined in `extract_code_graph` anyway
- [x] created tests for `qenv() |> eval_code |> get_code_attr("dependency")`
- [x] changed `extract_side_effects` and `extract_occurrence` so they work on an element, and they don't use `lapply`

### DIFF
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -82,10 +82,9 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
 
     pd <- utils::getParseData(current_call)
     pd <- normalize_pd(pd)
-    calls_pd <- extract_calls(pd)
+    call_pd <- extract_calls(pd)[[1]]
 
-    attr(current_code, "side_effects") <- extract_side_effects(calls_pd)[[1]]
-    attr(current_code, "occurrence") <- extract_occurrence(calls_pd)[[1]]
+    attr(current_code, "dependency") <- c(extract_side_effects(call_pd), extract_occurrence(call_pd))
     object@code <- c(object@code, list(current_code))
   }
 
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -79,6 +79,20 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
     }
 
     attr(current_code, "id") <- sample.int(.Machine$integer.max, size = 1)
+
+    # UNSURE if the removal of curly brackets is still needed.
+    tcode <- trimws(current_code)
+    if (any(grepl("^\\{.*\\}$", tcode))) {
+      tcode <- sub("^\\{(.*)\\}$", "\\1", tcode)
+    }
+
+    parsed_code <- parse(text = tcode, keep.source = TRUE)
+    pd <- utils::getParseData(parsed_code)
+    pd <- normalize_pd(pd)
+    calls_pd <- extract_calls(pd)
+
+    attr(current_code, "side_effects") <- extract_side_effects(calls_pd)[[1]]
+    attr(current_code, "occurrence") <- extract_occurrence(calls_pd)[[1]]
     object@code <- c(object@code, list(current_code))
   }
 
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -37,7 +37,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
 
   for (i in seq_along(code_split)) {
     current_code <- code_split[[i]]
-    current_call <- parse(text = current_code, keep.source = FALSE)
+    current_call <- parse(text = trimws(current_code), keep.source = TRUE)
 
     # Using withCallingHandlers to capture warnings and messages.
     # Using tryCatch to capture the error and abort further evaluation.
@@ -80,14 +80,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
 
     attr(current_code, "id") <- sample.int(.Machine$integer.max, size = 1)
 
-    # UNSURE if the removal of curly brackets is still needed.
-    tcode <- trimws(current_code)
-    if (any(grepl("^\\{.*\\}$", tcode))) {
-      tcode <- sub("^\\{(.*)\\}$", "\\1", tcode)
-    }
-
-    parsed_code <- parse(text = tcode, keep.source = TRUE)
-    pd <- utils::getParseData(parsed_code)
+    pd <- utils::getParseData(current_call)
     pd <- normalize_pd(pd)
     calls_pd <- extract_calls(pd)
 
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -37,7 +37,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
 
   for (i in seq_along(code_split)) {
     current_code <- code_split[[i]]
-    current_call <- parse(text = trimws(current_code), keep.source = TRUE)
+    current_call <- parse(text = current_code, keep.source = TRUE)
 
     # Using withCallingHandlers to capture warnings and messages.
     # Using tryCatch to capture the error and abort further evaluation.
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -79,12 +79,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
     }
 
     attr(current_code, "id") <- sample.int(.Machine$integer.max, size = 1)
-
-    pd <- utils::getParseData(current_call)
-    pd <- normalize_pd(pd)
-    call_pd <- extract_calls(pd)[[1]]
-
-    attr(current_code, "dependency") <- c(extract_side_effects(call_pd), extract_occurrence(call_pd))
+    attr(current_code, "dependency") <- extract_dependency(current_call)
     object@code <- c(object@code, list(current_code))
   }
 
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -463,31 +463,6 @@ split_code <- function(code) {
   gsub("^([[:space:]])*;(.+)$", "\\1\\2", new_code, perl = TRUE)
 }
 
-
-#' Reshape code to the list
-#'
-#' List will be divided by the calls. Each element of the list contains `id` and `dependency` attributes.
-#'
-#' @param code `character` with the code.
-#'
-#' @return list of `character`s of the length equal to the number of calls in `code`.
-#'
-#' @keywords internal
-#' @noRd
-code2list <- function(code) {
-  checkmate::assert_character(code, null.ok = TRUE)
-  if (length(code)) {
-    lapply(split_code(code), function(current_code) {
-      attr(current_code, "id") <- sample.int(.Machine$integer.max, 1)
-      parsed_code <- parse(text = trimws(current_code), keep.source = TRUE)
-      attr(current_code, "dependency") <- extract_dependency(parsed_code)
-      current_code
-    })
-  } else {
-    list(character(0))
-  }
-}
-
 #' @param parsed_code results of `parse(text = code, keep.source = TRUE` (parsed text)
 #' @keywords internal
 #' @noRd
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -53,7 +53,7 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
     }
   }
 
-  graph <- extract_code_graph(code)
+  graph <- lapply(code, attr, "dependency")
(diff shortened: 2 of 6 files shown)


R/qenv-eval_code.R
```diff
@@ -79,6 +79,20 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
...
 
     attr(current_code, "id") <- sample.int(.Machine$integer.max, size = 1)
+
+    # UNSURE if the removal of curly brackets is still needed.
+    tcode <- trimws(current_code)
+    if (any(grepl("^\\{.*\\}$", tcode))) {
+      tcode <- sub("^\\{(.*)\\}$", "\\1", tcode)
+    }
+
+    parsed_code <- parse(text = tcode, keep.source = TRUE)
```
gogonzo, line 89: there is `parsed_code` in this loop, let's use the first one crated on the top

m7pr: t@gogonzo the first parsed_code is based on `code` and second one on `tcode`. But I can move `tcode` creation to top, and only use `tcode`. Let's see if this doesn't break anything

m7pr:
Hey @gogonzo - if you have some spare time, this PR is ready for the first review.
It aims for transition of `code_graph` calculations from `get_code_dependency` into single calls in `eval_code`

R/utils-get_code_dependency.R
```diff
@@ -33,18 +33,11 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
...
-    tcode <- sub("^\\{(.*)\\}$", "\\1", tcode)
-  }
-
-  parsed_code <- parse(text = tcode, keep.source = TRUE)
+  parsed_code <- parse(text = trimws(code), keep.source = TRUE)
 
   pd <- utils::getParseData(parsed_code)
   pd <- normalize_pd(pd)
-  calls_pd <- extract_calls(pd)
+  calls_pd <- extract_calls(pd) # STILL NEEDED for check_names
```
gogonzo, line 40: It doesn't really look needed by `check_names`. IN the `if` below there are `unlist` anyway which means that instead of loop you can browse `pd` directly
```r
# instead
symbols <- unlist(lapply(calls_pd, function(call) call[call$token == "SYMBOL", "text"]))

# do
symbols <- pd[pd$token == "SYMBOL", "text"]
```

Also lines 36-39 should be moved inside of `if (check_names) {...}`

R/qenv-eval_code.R
```diff
@@ -79,6 +79,12 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
     }
 
     attr(current_code, "id") <- sample.int(.Machine$integer.max, size = 1)
+
+    pd <- utils::getParseData(current_call)
+    pd <- normalize_pd(pd)
+    call_pd <- extract_calls(pd)[[1]]
```
gogonzo, line 85: `extract_calls` while calls are already split?

R/utils-get_code_dependency.R
m7pr: Thanks, this simplified the code a bit. Moved it in this commit. 4ea2ae2be86b08bed72b41dd390f33c418bf4275

Do you think we could create a specific function for this check and move it out of `get_code_dependency` so it's shorter?

R/utils-get_code_dependency.R
```diff
@@ -33,23 +33,13 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
...
     # Detect if names are actually in code.
-    symbols <- unlist(lapply(calls_pd, function(call) call[call$token == "SYMBOL", "text"]))
+    parsed_code <- parse(text = trimws(code), keep.source = TRUE)
+    pd <- normalize_pd(utils::getParseData(parsed_code))
+    symbols <- pd[pd$token == "SYMBOL", "text"]
     if (any(pd$text == "assign")) {
+      calls_pd <- extract_calls(pd)
       assign_calls <- Filter(function(call) find_call(call, "assign"), calls_pd)
       ass_str <- unlist(lapply(assign_calls, function(call) call[call$token == "STR_CONST", "text"]))
       ass_str <- gsub("^['\"]|['\"]$", "", ass_str)
```
m7pr, line 45: Rethink if `calls_pd` is needed to figure out `STR_CONST` of objects used in `assign`. Maybe we can work on a full `pd`

R/qenv-eval_code.R
m7pr: `extract_occurrence` assume that pd has a specific order of rows, that gets changed when you apply `extract_calls(pd)` on `pd`.

I can move `call_pd <- extract_calls(pd)[[1]]` inside `extract_occurrence`. Or is it ok to keep in eval_code?

gogonzo: Let me ask another question - why to reorder `pd`?

m7pr: `extract_occurrence` assumes object names (`SYMBOL`) appear after `<-` (`LEFT_ASSIGN`/`ASSIGN`). 

`extract_calls` reorders `pd` in this way.

```r
code <- "a<-1"
parsed_code <- parse(text = code, keep.source = TRUE)
pd <- utils::getParseData(parsed_code)
> pd
  line1 col1 line2 col2 id parent       token terminal text
7     1    1     1    4  7      0        expr    FALSE     
1     1    1     1    1  1      3      SYMBOL     TRUE    a
3     1    1     1    1  3      7        expr    FALSE     
2     1    2     1    3  2      7 LEFT_ASSIGN     TRUE   <-
4     1    4     1    4  4      5   NUM_CONST     TRUE    1
5     1    4     1    4  5      7        expr    FALSE     
> extract_calls(pd)
  line1 col1 line2 col2 id parent       token terminal text
7     1    1     1    4  7      0        expr    FALSE     
3     1    1     1    1  3      7        expr    FALSE     
2     1    2     1    3  2      7 LEFT_ASSIGN     TRUE   <-
5     1    4     1    4  5      7        expr    FALSE     
1     1    1     1    1  1      3      SYMBOL     TRUE    a
4     1    4     1    4  4      5   NUM_CONST     TRUE    1
```

If we don't want to use `extract_calls` we need to revisit and rewrite `extract_occurrence`. I am not sure we will invent rules to figure out dependencies for a non-sorted `pd`.

So right now this is a matter of whether we 
- use `extract_calls` on `pd` and before we apply `pd` inside `extract_occurrence` 
- or whether we put `extract_calls` inside `extract_occurrence` 
- or whether we rewrite `extract_occurrence`.

`extract_occurrence` is a pretty big beast : P

m7pr: Maybe we can simplify `extract_calls` part so that it only reorders

m7pr: I just played with what parts of `extract_calls` are needed so that `extract_occurrence` works and below are the needed parts:

```r
# reordering basen on parent-children relation
parent_ids <- pd[pd$parent == 0 & (pd$token !e= "COMMENT" | grepl("@linksto", pd$text, fixed = TRUE)), "id"]
pd_order <- do.call(rbind, lapply(parent_ids, function(parent) rbind(pd[pd$id == parent, ], get_children(pd, parent))))
# filtering or edge cases
if (!is.null(pd_order) && !(nrow(pd_order) == 1 && call$token == "';'")) {
    # fixing assignment arrows
    pd_order <- fix_arrows(list(pd_order))[[1]]
    attr(current_code, "dependency") <- c(extract_side_effects(pd_order), extract_occurrence(pd_order))
}
```

So this is basically all of `extract_calls` without `fix_shifted_comments` that is skipped if there is only 1 call.


If `extract_calls` would be renamed to `reorder_and_clean_calls` then I suppose usage of this function is totally justified in here. We can always put this part in `extract_occurrence` but it's gonna be repeated in `extract_calls` and `extract_occurrence`

m7pr:
Hey @gogonzo I also needed to change the way we create `teal.data::teal_data()`
a0d4890b7816b614ba353be8db2175ccfb6c39ed
so that it's `@code` field corresponds to the changes made in here. 

There is a bit code from teal.code for unexported functions. Should we export something in teal.code, so that it could be reused in teal.data?

R/qenv-eval_code.R
```diff
@@ -37,7 +37,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
 
   for (i in seq_along(code_split)) {
     current_code <- code_split[[i]]
-    current_call <- parse(text = current_code, keep.source = FALSE)
+    current_call <- parse(text = trimws(current_code), keep.source = TRUE)
```
gogonzo, line 40: Just curious, why white spaces are problematic?

R/qenv-eval_code.R
```diff
@@ -79,6 +79,7 @@ setMethod("eval_code", signature = c("qenv", "character"), function(object, code
     }
 
     attr(current_code, "id") <- sample.int(.Machine$integer.max, size = 1)
+    attr(current_code, "dependency") <- extract_dependency(current_call)
```
gogonzo, line 82: This loop looks nice now. `extract_dependency` is a good move 👍

R/utils-get_code_dependency.R
```diff
@@ -33,23 +33,13 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
...
-  parsed_code <- parse(text = tcode, keep.source = TRUE)
-
-  pd <- utils::getParseData(parsed_code)
-  pd <- normalize_pd(pd)
-  calls_pd <- extract_calls(pd)
-
   if (check_names) {
     # Detect if names are actually in code.
-    symbols <- unlist(lapply(calls_pd, function(call) call[call$token == "SYMBOL", "text"]))
+    parsed_code <- parse(text = trimws(code), keep.source = TRUE)
```
gogonzo, line 38: Looks like in this if you try to obtain informations which are already in the `lapply(code, attr, "depedency")`

tests/testthat/test-qenv_eval_code.R
```diff
@@ -129,3 +129,21 @@ testthat::test_that("eval_code returns a qenv object with empty messages and war
...
+    lapply(q2@code, attr, "dependency"),
+    list(
+      c("x", "<-"),
+      c("iris_data", "<-", "head", "iris"),
+      c("x", "<-", "nrow", "iris_data")
+    )
+  )
+
+  q3 <- eval_code(qenv(), c("library(survival)", "library(utils)", "x <- 5"))
+  lapply(q3@code, attr, "dependency")
```
gogonzo, line 148: three tests in one. Please split, the last one has no expectation

R/utils-get_code_dependency.R
m7pr: Thanks for a fresh look! Will investigate if this could be simplified based on `lapply(code, attr, "depedency")`

R/qenv-eval_code.R
m7pr: I think that's a remaining of when we also removed `{` (brackets) at the beginning and the end of the call. So we first removed whitespaces, and then brackets. I'll try to see if this `trimws` is still needed here. Probably not

R/utils-get_code_dependency.R
m7pr: Simplified in here a8016bc4cf1ee494eae795ef1c722e0db89ed42a

R/qenv-eval_code.R
m7pr: Removed! 8d3649e6b676dd9c1e4c2dc1016138c13d1be557

tests/testthat/test-qenv_eval_code.R
m7pr: Split them in here 90fab4f4deef42e8d63560bef7d6149b715fac0d

R/utils-get_code_dependency.R
```diff
@@ -511,3 +457,12 @@ split_code <- function(code) {
...
   gsub("^([[:space:]])*;(.+)$", "\\1\\2", new_code, perl = TRUE)
 }
+
+#' @param parsed_code results of `parse(text = code, keep.source = TRUE` (parsed text)
+#' @keywords internal
+#' @noRd
+extract_dependency <- function(parsed_code) {
+  pd <- normalize_pd(utils::getParseData(parsed_code))
+  call_pd <- extract_calls(pd)[[1]]
+  c(extract_side_effects(call_pd), extract_occurrence(call_pd))
```
gogonzo, line 467:
```suggestion
  reordered_pd <- extract_calls(pd)[[1]] # because ...
  c(extract_side_effects(reordered_pd), extract_occurrence(reordered_pd))
```

gogonzo (approved):
All good here 👍 Please see the comment below

PR #225, opened 2024-11-08
Remove magrittr dependency

### LINKED ISSUE #224: [Question]: Can we remove `magrittr` from Suggests?
After https://github.com/insightsengineering/teal.code/pull/216/ will be merged, looks like `magrittr` is not used anywhere in the whole package. Should we just remove from Suggests in DESCRIPTION?

llrs-roche:
# Pull Request

Fixes #224 

- Depends on #216 

Checks with --as-cran ran clean.
I didn't find any instance where %>% was used in the package and as it is suggested I don't think it will be needed to check teal.gallery or other packages. But no reverse dependencies were done yet.


m7pr (changes requested):
There is more places where you need to remove magrittr :P

image

m7pr (approved):

m7pr:
Hey @llrs-roche good job. Let's wait till Monday to get this merged - unsure if this will break the current changes we are doing on other PR in teal.code

PR #231, opened 2024-11-12
Fix join on some platforms

### LINKED ISSUE #230: [Bug]: test failure on windows and linux
### What happened?

https://github.com/r-universe/pharmaverse/actions/runs/11777449384/job/32802007063

gogonzo:
closes #230 
Probably `union` on some platform doesn't take attributes into account (due to as.vector call). Also union/unique is not recommended for lists

`?unique` 
image

### DIFF
--- a/R/qenv-c.R
+++ b/R/qenv-c.R
@@ -89,8 +89,8 @@ c.qenv <- function(...) {
         stop(join_validation)
       }
 
-      x_id <- vapply(x@code, attr, which = "id", integer(1))
-      y_id <- vapply(y@code, attr, which = "id", integer(1))
+      x_id <- get_code_attr(x, "id")
+      y_id <- get_code_attr(y, "id")
       x@code <- c(x@code, y@code[!y_id %in% x_id])
 
       # insert (and overwrite) objects from y to x
--- a/R/qenv-c.R
+++ b/R/qenv-c.R
@@ -89,7 +89,9 @@ c.qenv <- function(...) {
         stop(join_validation)
       }
 
-      x@code <- union(x@code, y@code)
+      x_id <- vapply(x@code, attr, which = "id", integer(1))
+      y_id <- vapply(y@code, attr, which = "id", integer(1))
+      x@code <- c(x@code, y@code[!y_id %in% x_id])
 
       # insert (and overwrite) objects from y to x
       x@.xData <- rlang::env_clone(x@.xData, parent = parent.env(.GlobalEnv))


R/qenv-c.R
```diff
@@ -89,7 +89,9 @@ c.qenv <- function(...) {
         stop(join_validation)
       }
 
-      x@code <- union(x@code, y@code)
+      x_id <- vapply(x@code, attr, which = "id", integer(1))
+      y_id <- vapply(y@code, attr, which = "id", integer(1))
```
m7pr, line 93: What about below? We have those used in `.check_joinable`. If you don't like the suggestion, let's also unify there and let's use `vapply` there. In this case the in-build wrapper is shorter but it's not explicitly visible that integers will be removed.

```suggestion
      x_id <- get_code_attr(x, "id")
      y_id <- get_code_attr(y, "id")
```

m7pr:
@gogonzo looks good, left a small comment

R/qenv-c.R
gogonzo: It makes sense. I was thinking even about something else. If `@code` was a named list with "id" as names, then `modifyList(x@code, y@code)` would be needed.

m7pr: Hmm, then `id` would be a name of the element. So you could also run `names(q@code)` to see ids. The only drawback (that is minor) is the fact those ids would be converted to characters? Not a big deal.

One more benefit is that you could subset the code by ids `q@code[c('id1', 'id2')]` and `q@code[['id1']]`. Maybe it's worth it

gogonzo: I'll try and you'll see if its worth it 👍

m7pr: One day I was thinking also about using `hash`es instead of `ids` for a call, but those would not be unique if the call is repeated. However we could have `hash`es as attributes, so that we can identify unique calls and calls that appear twice. It's not a big improvement and doesn't solve a lot. Just an extra metadata

gogonzo: But hash would be an id right? Maybe there would be a different naming convention, but process would look exactly the same. isn't it?

m7pr: Hash would be an attribute, the same way as `warning` and `message`. But this is not a must to have. You can always calculate hashes on your own when you need them. Just `lapply(q@code, rlang::hash)`

gogonzo: Yes, but as you noted before - calculated on the same looking code would return the same value. We can't calculate hash afterwards. It must be assigned to the specific evaluated code. Otherwise there will be a join problem.

m7pr: It was just a thought. No need to implement that : )

gogonzo: Yup. But that was a valid suggestion. I'm resolving this conversation, feel free to unresolve.

m7pr (approved):

PR #232, opened 2024-11-13
id as names

### LINKED ISSUE #230: [Bug]: test failure on windows and linux

gogonzo:
closes #230 
alternative to #231 

instead of `id` being an attribute of the `@code` elements it is now a name of the `@code` elements.
Why it fits to the `names(@code)`?
- id is a single value for each code element
- id must be unique across all elements


m7pr:
Hey, this looks good on a first glance. I will make a through review once I'm done with decorator PR on teal

tests/testthat/test-qenv_eval_code.R
```diff
@@ -127,7 +127,7 @@ testthat::test_that("comments alone are pasted to the next/following call elemen
   code <- c("x <- 5", "# comment", "y <- 6")
   q <- eval_code(qenv(), code)
   testthat::expect_identical(
-    unlist(q@code)[2],
+    as.character(q@code)[[2]],
```
m7pr, line 130: `as.character` make it already a vector
```suggestion
    as.character(q@code)[2],
```

m7pr:
A minor comment not related to this PR, but a food for thought

> id must be unique across all elements

There is a slight chance that the id will not be the same if we do `sample.int(.Machine$integer.max, size = 1)` but the chance is small (1/.Machine$integer.max). To make the chance be 0% I would provide hashes of the time of the evaluation

```r
> rlang::hash(Sys.time())
[1] "bda544bca9e09b34089d8a1bd5788fa7"
> Sys.time()
[1] "2024-11-14 09:31:26 CET"
```

m7pr (approved):
Good to go, left one comment in tests.
This is great enhancement. I think it will be easier to work with names, instead of attributes

gogonzo:
> rlang::hash(Sys.time())

This is not enough. There are few evaluation within a second, on a simple apps and fast machines you'll get duplicates.

Issue #235, opened 2024-11-20
[Bug]: [WARN] In ‘sym_cond > ass_cond’: longer object length is not a multiple of shorter object length


m7pr:
Potentially caused by `teal.code:::get_code_dependency`
### What happened?

When running below on this branch https://github.com/insightsengineering/teal.modules.general/pull/795

```r
devtools::load_all("../teal")
devtools::load_all(".")
footnote_regression <- teal_transform_module(
  server = make_teal_transform_server(expression(
    plot <- plot + labs(caption = deparse(summary(fit)[[1]]))
  ))
)

data <- teal_data()
data <- within(data, {
  require(nestcolor)
  ADSL <- rADSL
})
join_keys(data) <- default_cdisc_join_keys[names(data)]

app <- init(
  data = data,
  modules = modules(
    tm_a_regression(
      label = "Regression",
      response = data_extract_spec(
        dataname = "ADSL",
        select = select_spec(
          label = "Select variable:",
          choices = "BMRKR1",
          selected = "BMRKR1",
          multiple = FALSE,
          fixed = TRUE
        )
      ),
      regressor = data_extract_spec(
        dataname = "ADSL",
        select = select_spec(
          label = "Select variables:",
          choices = variable_choices(data[["ADSL"]], c("AGE", "SEX", "RACE")),
          selected = "AGE",
          multiple = TRUE,
          fixed = FALSE
        )
      ),
      decorators = list(footnote_regression)
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}


```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #236, opened 2024-11-27
Fix `‘sym_cond > ass_cond’: longer object `

m7pr:
Fixes https://github.com/insightsengineering/teal.code/issues/235

There was a warning thrown by the `get_code_dependency` that is now not visible 
+ the code extraction works for `for` loops.

Added 2 tests to prove that.

### DIFF
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -274,6 +274,7 @@ extract_occurrence <- function(pd) {
   # What occurs in a function body is not tracked.
   x <- pd[!is_in_function(pd), ]
   sym_cond <- which(x$token %in% c("SPECIAL", "SYMBOL", "SYMBOL_FUNCTION_CALL"))
+  sym_fc_cond <- which(x$token == "SYMBOL_FUNCTION_CALL")
 
   if (length(sym_cond) == 0) {
     return(character(0L))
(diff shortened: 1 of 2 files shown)


R/utils-get_code_dependency.R
```diff
@@ -287,28 +288,30 @@ extract_occurrence <- function(pd) {
...
+  if (!length(assign_cond)) {
     return(c("<-", unique(x[sym_cond, "text"])))
   }
 
-  sym_cond <- sym_cond[sym_cond > ass_cond] # NOTE 1
+  # For cases like 'eval(expression(c <- b + 2))' removes 'eval(expression('.
+  sym_cond <- setdiff(
+    sym_cond,
+    sym_cond[sym_cond < min(assign_cond) & sym_cond %in% sym_fc_cond]
+  )
```
averissimo, line 300: Can the same be achieved with

```suggestion
  sym_cond <- sym_cond[!(sym_cond < min(assign_cond) & sym_cond %in% sym_fc_cond)]
```

tests/testthat/test-qenv_get_code.R
```diff
@@ -596,6 +596,54 @@ testthat::test_that("detects occurrence of a function definition with a @linksto
...
+
+testthat::test_that("objects in for loop are extracted if passed as separate calls", {
+  q <- within(qenv(), {
+    a <- 1
+    b <- 2
+  }) |> within({
+    for (x in c(1, 2)) {
+      b <- a
+      b <- b + a + 1
+      b + 3 -> b
```
averissimo, line 629:
```suggestion
      b + 3 -> b # nolint: assignment.
```

R/utils-get_code_dependency.R
averissimo: (optional)

This is a question of personal opinion on readability. 

Feel free to keep it the way you consider is best / consistent with `teal.code`!

PR #239, opened 2024-12-18
datanames in vignettes

gogonzo:
Refers to #338 

- Modified a documentation a little bit to emphasize inheritance from `environment`.
- @averissimo I'm thinking to move description of `eval_code`, `within`, `get_code` away from details of `qenv` to a new section "Interacting with `qenv`". 
- I'm thinking also to use a term "`qenv` environment" instead of "`qenv` object" to emphasize it is an environment.


R/qenv-eval_code.R
```diff
@@ -1,16 +1,21 @@
...
 #' @details
+#'
 #' `eval_code()` evaluates given code in the `qenv` environment and appends it to the `code` slot.
 #' Thus, if the `qenv` had been instantiated empty, contents of the environment are always a result of the stored code.
+#' The `qenv` object is immutable, even though it inherits from the environment class, which is typically mutable by
+#' design. This means that each code evaluation does not modify the original `qenv` object directly. Instead, every
+#' code evaluation creates a new `qenv` object that reflects the result of the changes, leaving the original object
+#' unchanged.
 #'
 #' @param object (`qenv`)
```
averissimo, line 12: :+1: to using refer `qenv` as "environment" instead of "object"

As a side note, I was thinking if we should deprecate this argument name in favor of something else:
-  `envir` is too generic as we  don't accept base environments
- I would like `qenv` but it might be too specific
- `data` would make it similar to `within` arguments, but not very descriptive

Keeping as is feels the best approach

<\side note>

averissimo: Crazy idea that would allow `eval_code` to accept base environments and convert them to qenv, but I cannot justify it as:

1. We shouldn't support the same in `within` as that's a base-level generic/object pair
2. It also doesn't make sense in terms of using it
    - Just as it doesn't make sense calling `eval_code`/`within` after `teal_data(iris = iris, mtcars = mtcars)`
        - Although in this case it's possible.

R/qenv-constructor.R
```diff
@@ -11,8 +11,23 @@
...
+#'
+#' A `qenv` inherits from the `environment` class, behaves like an environment, and has the
+#' following characteristics:
+#'
+#' - `qenv` environment is locked, and data modification is only possible through the `eval_code()`
+#' and `within()` functions.
+#' - It stores metadata about the code used to create the data.
+#' - Is immutable which means that each code evaluation does not modify the original `qenv`
+#' environment directly. See the following code:
+#'   ```
```
averissimo, line 26: Minor markdown details

```suggestion
#'
#' - It stores metadata about the code used to create the data.
#' - Is immutable which means that each code evaluation does not modify the original `qenv`
#' environment directly. See the following code:
#'
#'   ```
```

averissimo (commented):
> - @averissimo I'm thinking to move description of `eval_code`, `within`, `get_code` away from details of `qenv` to a new section "Interacting with `qenv`". 

Keeping it in `@details` makes it consistent with other R documentation and our own.

But I'm not opposed to this change if it improves the readability

R/qenv-eval_code.R
gogonzo: > As a side note, I was thinking if we should deprecate this argument name in favor of something else:

I'd add a `x` to this list also

> Crazy idea that would allow eval_code to accept base environments and convert them to qenv, but I cannot justify it as

I like where it goes but at this moment we can be pretty confident with the design. However, replacing `teal_data(a = a, b = b)` constructor with an environment which could be supported by `eval_code` sounds also nice.

> Just as it doesn't make sense calling eval_code/within after teal_data(iris = iris, mtcars = mtcars)

It makes some sense. You can initialize `teal_data` object with some code and data as `!verified` and it could still be a reproducible code and data.

gogonzo:
> Keeping it in @details makes it consistent with other R documentation and our own.
> But I'm not opposed to this change if it improves the readability

Yup, I think it is fine now. It was my first thought with the `@details` but right now sections looks better.

R/qenv-constructor.R
gogonzo: I don't understand what this change suppose to do

averissimo: It was adding empty line before list and before chunk of code, I see it renders great now, ignore it

R/qenv-within.R
```diff
@@ -1,22 +1,17 @@
...
-#'
 #' @details
-#' `within()` is a convenience function for evaluating inline code inside the environment of a `qenv`.
-#' It is a method for the `base` generic that wraps `eval_code` to provide a simplified way of passing code.
-#' `within` accepts only inline expressions (both simple and compound) and allows for injecting values into `expr`
-#' through the `...` argument:
-#' as `name:value` pairs are passed to `...`, `name` in `expr` will be replaced with `value`.
+#' `within()` is a convenience method that wraps `eval_code` to provide a simplified way of passing expression.
+#' `within` accepts only inline expressions (both simple and compound) and allows to substitute `expr`
+#' with the `...` argument values.
```
averissimo, line 4:
```suggestion
#' with `...` named argument values.
```

R/qenv-within.R
```diff
@@ -1,22 +1,17 @@
...
 #'
 #' @section Using language objects with `within`:
 #' Passing language objects to `expr` is generally not intended but can be achieved with `do.call`.
 #' Only single `expression`s will work and substitution is not available. See examples.
 #'
 #' @param data (`qenv`)
 #' @param expr (`expression`) to evaluate. Must be inline code, see `Using language objects...`
-#' @param ... see `Details`
+#' @param ... (`named`) argument value will substitute a symbol in the `expr` matched by the name.
```
averissimo, line 12: `named` is not a data type

WDYT of breaking standard here:

```suggestion
#' @param ... named argument value will substitute a symbol in the `expr` matched by the name.
```

R/qenv-eval_code.R
averissimo: > You can initialize teal_data object with some code and data as !verified and it could still be a reproducible code and data.

You got a point here, I see people doing this for computational heavy operations that would be avoided, or if the code/values come from a different "system"

R/qenv-within.R
```diff
@@ -1,22 +1,17 @@
...
+#' For practical examples see the #usage.
```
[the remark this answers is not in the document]
averissimo: at least in rstudio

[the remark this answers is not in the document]
gogonzo: Yeah, I meant For practical usage see the examples - changed but I don't know how to link Examples section. I don't think it is necessary but would be nice

DESCRIPTION
```diff
@@ -28,13 +28,13 @@ Depends:
     R (>= 4.0)
 Imports:
     checkmate (>= 2.1.0),
+    cli (>= 3.4.0),
```
averissimo, line 31: Needs to be added to `Config/Needs/verdepcheck: mllg/checkmate, r-lib/lifecycle, r-lib/rlang,
Config/Needs/verdepcheck` as well (and `.pre-commit-config.yaml`

gogonzo: added 👍

R/qenv-show.R
```diff
@@ -16,5 +16,26 @@
...
+  locked <- if (environmentIsLocked(env)) {
+    " [L]"
+  } else {
+    ""
+  }
+  parent <- sprintf("Parent: <environment: %s>", rlang::env_label(rlang::env_parent(env)))
+  cat(cli::style_bold(paste0(header, locked)), sep = "\n")
+  cat(parent, "\n")
+
+  shown <- ls(object)
```
averissimo, line 30: image
image


```suggestion
  shown <- vapply(rlang::syms(ls(object)), deparse, character(1L), backtick = TRUE)
```

averissimo: Oh wait, this creates a problem with class after

R/qenv-show.R
```diff
@@ -16,5 +16,26 @@
...
+  lapply(shown, function(x) {
+    cat(sprintf("- %s: [%s]\n", x, class(object[[x]])[1]))
```
averissimo, line 32:
```suggestion
    cat(sprintf("- %s: [%s]\n", deparse(rlang::sym(x), backtick = TRUE), class(object[[x]])[1]))
```

R/qenv-show.R
gogonzo: true

R/qenv-show.R
```diff
@@ -16,5 +16,32 @@
...
+    " [L]"
+  } else {
+    ""
+  }
+  parent <- sprintf("Parent: <environment: %s>", rlang::env_label(rlang::env_parent(env)))
+  cat(cli::style_bold(paste0(header, locked)), sep = "\n")
+  cat(parent, "\n")
+
+  shown <- ls(object)
+  lapply(shown, function(x) {
```
averissimo, line 31: We don't necessarly want to keep 1-1 match, with `rlang::env_print()` (the [L] indicator for each binding seems overkill)

I do like the `Bindings:` bolded header before the variable names.

It's self-explanatory, but it gives structure to the print. I would even consider adding it to the hidden objects.

image


```suggestion
  if (length(shown > 0L)) cat(cli::style_bold("Bindings:\n"))
  lapply(shown, function(x) {
```

gogonzo: **Bindings** yes, but **Hidden Bindings:** not IMO. I thought about having section hidden: before and decided to not do so. I think dot-prefixed greyed out is a sufficient information

averissimo: Sounds good :+1:

averissimo (commented):
Great job with the custom print method, the grayed out names is a brilliant touch

I think it works great with most cases I can think of. It doesn't show the right class with `rlang::missing_arg()`, but that is a very special case

``` r
pkgload::load_all("teal.code", export_all = TRUE)
weird_classes <- qenv() |> 
  within({
    a <- 2
    b <- 5 + 1i
    c <- 10L
    .d_e <- new.env()
    data("miniACC", package = "MultiAssayExperiment", envir = .d_e)
    .d <- .d_e$miniACC
    e <- structure(list(1), class = c("yada", "bla"))
    .f <- iris
    ` a very !weird name` <- R6::R6Class("A", public = list(a = 1))
    an_r6_instance <- R6::R6Class("A Class", public = list(a = 1))$new()
    g <- structure(list(1), class = "override")
    class(g) <- 2
    h <- function(x) 1234
    x <- rlang::missing_arg()
    y <- quote(foo)
    z <- expression(1+1)
  })
weird_classes
#> <environment: 0x6132e7c2c728> [L]
#> Parent: <environment: package:MultiAssayExperiment> 
#> - ` a very !weird name`: [R6ClassGenerator]
#> - a: [numeric]
#> - an_r6_instance: [A Class]
#> - b: [complex]
#> - c: [integer]
#> - e: [yada]
#> - g: [2]
#> - h: [function]
#> - x: [name]
#> - y: [name]
#> - z: [expression]
#> - .d: [MultiAssayExperiment]
#> - .d_e: [environment]
#> - .f: [data.frame]
rlang::env_print(as.environment(weird_classes))
#> <environment: 0x6132e7c2c728> [L]
#> Parent: <environment: package:MultiAssayExperiment>
#> Bindings:
#> • x: <missing> [L]
#> • y: <sym> [L]
#> • z: <expression> [L]
#> • a: <dbl> [L]
#> • b: <cpl> [L]
#> • c: <int> [L]
#> • e: <yada> [L]
#> • an_r6_instance: <A Class> [L]
#> • g: <2> [L]
#> • h: <fn> [L]
#> • ` a very !weird name`: <R6ClssGn> [L]
#> • .d_e: <env> [L]
#> • .d: <MltAssyE> [L]
#> • .f: <df[,5]> [L]
```

<sup>Created on 2024-12-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/qenv-show.R
```diff
@@ -16,5 +16,28 @@
...
+        "- %s: [%s]\n",
+        deparse(rlang::sym(x), backtick = TRUE),
+        class(object[[x]])[1]
+      )
+    )
+  })
+
+  hidden <- setdiff(ls(object, all.names = TRUE), shown)
+  lapply(hidden, function(x) {
+    cat(cli::style_blurred(sprintf("- %s: [%s]\n", x, class(object[[x]])[1])))
```
averissimo, line 39: Applying same backticked naming here

```suggestion
    cat(
      cli::style_blurred(
        sprintf(
          "- %s: [%s]\n",
          deparse(rlang::sym(x), backtick = TRUE),
          class(object[[x]])[1]
        )
      )
    )
```

R/qenv-show.R
```diff
@@ -16,5 +16,28 @@
...
+  cat(parent, "\n")
+
+  shown <- ls(object)
+  if (length(shown > 0L)) cat(cli::style_bold("Bindings:\n"))
+  lapply(shown, function(x) {
+    cat(
+      sprintf(
+        "- %s: [%s]\n",
+        deparse(rlang::sym(x), backtick = TRUE),
+        class(object[[x]])[1]
```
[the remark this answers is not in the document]
gogonzo: I'm not a fan of changing original class names neither using extra dependency for it

averissimo (approved):
It looks  really good, repeated docs seem to be unified

PR #244, opened 2025-02-04
Fix infinite loop

gogonzo:
closes https://github.com/insightsengineering/teal/issues/1474

Problem happens when two or more symbols occur in the same time on rhs and lhs. Due to this, `l` depends on `class` and `class` depends on `l`.

```r
  code <- "l <- list(a = 1, b = 2)
    class(l) <- c('new class', class(l))" # l depends on class and class depends on l
  q <- eval_code(qenv(), code)
  get_code(q, names = "l")
```

@donyunardi This was a serious bug and will require a new release. We can wait a little to the day when `teal` is going to be submitted to CRAN.

### DIFF
--- a/tests/testthat/test-qenv_get_code.R
+++ b/tests/testthat/test-qenv_get_code.R
@@ -951,3 +951,10 @@ testthat::test_that("original formatting and comments are preserved when express
   expr <- parse(text = code, keep.source = TRUE)
   testthat::expect_identical(get_code(eval_code(qenv(), expr)), code)
 })
+
+testthat::test_that("extracting code doesn't fail when lhs contains two or more symbols occurring in lhs", {
+  code <- "l <- list(a = 1, b = 2)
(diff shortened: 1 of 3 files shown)


tests/testthat/test-qenv_get_code.R
```diff
@@ -951,3 +951,10 @@ testthat::test_that("original formatting and comments are preserved when express
   expr <- parse(text = code, keep.source = TRUE)
   testthat::expect_identical(get_code(eval_code(qenv(), expr)), code)
 })
+
+testthat::test_that("extracting code doesn't fail when lhs contains two or more symbols occurring in lhs", {
```
chlebowa, line 955:
```suggestion
testthat::test_that("extracting code doesn't fail when lhs contains two or more symbols occurring in rhs", {
```

m7pr (approved):
well done @gogonzo 

thanks @chlebowa for raising

PR #254, opened 2025-06-03
Improvements (i hope so)

gogonzo:
Code balance is + because I've added one test ;]

### DIFF
--- a/R/qenv-within.R
+++ b/R/qenv-within.R
@@ -48,20 +48,13 @@
 #' @export
 #'
 within.qenv <- function(data, expr, ...) {
-  expr <- substitute(expr)
+  expr <- as.expression(substitute(expr))
   extras <- list(...)
 
-  # Add braces for consistency.
-  if (!identical(as.list(expr)[[1L]], as.symbol("{"))) {
-    expr <- call("{", expr)
-  }
-
-  calls <- as.list(expr)[-1]
-
   # Inject extra values into expressions.
-  calls <- lapply(calls, function(x) do.call(substitute, list(x, env = extras)))
+  calls <- lapply(expr, function(x) do.call(substitute, list(x, env = extras)))
 
-  eval_code(object = data, code = as.expression(calls))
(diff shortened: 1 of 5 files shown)


R/qenv-within.R
```diff
@@ -48,20 +48,13 @@
 #' @export
 #'
 within.qenv <- function(data, expr, ...) {
```
averissimo, line 50: Should we check for named dots?

``` r
#> ℹ Loading teal.code
qenv() |> within(aa <- 1+1, TRUE, FALSE)
#> <environment: 0x5b1c29313ae0> 🔒 
#> Parent: <environment: devtools_shims> 
#> Bindings:
#> - aa: [numeric]
```

<sup>Created on 2025-06-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/qenv-within.R
```diff
@@ -48,20 +48,13 @@
...
-  expr <- substitute(expr)
+  expr <- as.expression(substitute(expr))
   extras <- list(...)
```
averissimo, line 52: I'm not sure about both:
- Substituting `...` in `expr`
- Passing `...` to `eval_code()`

Maybe we should restrict `...` to be ignore the formals of `eval_code`

``` r
pkgload::load_all("/home/averissimo/work/roche🟦/packages📦/teal.code")
#> ℹ Loading teal.code
qenv() |> within(1+1, object = TRUE)
#> Error in eval_code(object = data, code = as.expression(calls), ...): formal argument "object" matched by multiple actual arguments
```

<sup>Created on 2025-06-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

```suggestion
  extras <- list(...)
  extras <- extras[setdiff(names(extras), setdiff(names(formals(eval_code)), c("object", "data")))]
```

averissimo (commented):
Awesome, I have one comment about passing the arguments forward

R/qenv-within.R
averissimo: Awesome

PR #257, opened 2025-06-12
Keeps last value of `eval_code()`/`within()`

averissimo:
New feature for `teal.code` that keeps the last value of the code evaluation.

In line with R's `.Last.value` (see [doc](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/Last.value))

image

### DIFF
--- a/R/qenv-eval_code.R
+++ b/R/qenv-eval_code.R
@@ -13,6 +13,8 @@
 #'
 #' @return
 #' `qenv` environment with `code/expr` evaluated or `qenv.error` if evaluation fails.
+#' The environment contains an attribute called `".Last.value"` which is the last evaluated value,
+#' similarly to [base::.Last.value].
 #'
 #' @examples
 #' # evaluate code in qenv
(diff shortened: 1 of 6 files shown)


R/qenv-eval_code.R
```diff
@@ -61,7 +63,8 @@ setMethod("eval_code", signature = c(object = "qenv.error"), function(object, co
     x <- withCallingHandlers(
       tryCatch(
         {
-          eval(current_call, envir = object@.xData)
+          .Last.value <- eval(current_call, envir = object@.xData)
+          attr(object@.xData, ".Last.value") <- .Last.value
```
gogonzo, line 67: Comparing to base - `.Last.value` isn't an attribute but an object of the .GlobalEnv

gogonzo: Hmmm, even worse. `.Last.value` is an object in `base` and not assigned to any environment :/

averissimo:
@gogonzo Should we pause/discard this PR given your new discovery late yesterday and our call today in favor of using `evaluate::evaluate()` with custom handler.

This could live as its own feature though.

averissimo:
Closed in favor of #259

PR #258, opened 2025-06-19
Use evaluate

gogonzo:
`eval_code` caches now outputs of the code in the `attr(<code line>, "outputs")`. See the tests of `get_outputs` to see how it works.

### DIFF
--- a/vignettes/qenv.Rmd
+++ b/vignettes/qenv.Rmd
@@ -14,6 +14,7 @@ A `qenv` inherits from the `environment` class, behaves like an environment, and
 
 - It inherits from the environment and methods such as `$`, `get`, `ls`, `as.list()` work out of the box.
 - `qenv` is a locked environment, and data modification is only possible through the `eval_code` and `within` functions.
+- It stores printed and plotted outputs (see `get_outputs`).
 - It stores metadata about the code used to create the data (see `get_code`).
 - It supports slicing by `[`.
 - It is immutable which means that each code evaluation does not modify the original `qenv` environment directly.
@@ -56,13 +57,16 @@ The same result can be achieved with the `within` method.
 ```{r}
 q2 <- within(my_qenv, y <- x * 2)
 q2 <- within(q2, z <- y * 2)
+q2 <- within(q2, plot(z))
 print(q2)
 ```
 
-To extract objects from a `qenv`, use `[[`; this is particularly useful for displaying them in a `shiny` app. You can retrieve the code used to generate the `qenv` using the `get_code()` function.
+To extract specific object from a `qenv`'s environment, use `[[`. To extract an output of a `print` or `plot` functions, use `get_outputs()` to get a `list()` of outputs captured by `qenv`. These functions are particularly useful for displaying them in a `shiny` app. You can retrieve the code used to generate the `qenv` using the `get_code()` function.
 
 ```{r}
 print(q2[["y"]])
+
+print(get_outputs(q2)[[1]])
 
 cat(get_code(q2))
 ```
@@ -126,7 +130,8 @@ The feasibility of joining `qenv` objects hinges on the contents of the environm
 
 ### Warnings and messages in `qenv` objects
 
-In cases where warnings or messages arise while evaluating code within a `qenv` environment, these are captured and stored within the `qenv` object. Access these messages and warnings using below
+In cases where warnings or messages arise while evaluating code within a `qenv` environment, these are captured and stored within the `qenv` object. Access these messages and warnings using `get_messages()` and `get_warnings()` functions as shown below.
+ 
 
 ```{r}
 q_message <- eval_code(qenv(), quote(message("this is a message")))
@@ -163,13 +168,13 @@ server <- function(input, output, session) {
   # create a qenv containing the reproducible output
   output_q <- reactive({
     req(input$option)
-    eval_code(
+    within(
       data_q,
-      bquote(p <- hist(iris_data[, .(input$option)]))
+      p <- hist(iris_data[, .(input$option)])
     )
   })
 
-  # display output
+  # display plot output
   output$plot <- renderPlot(output_q()[["p"]])
   # display code
   output$rcode <- renderText(get_code(output_q()))


R/qenv-eval_code.R
```diff
@@ -53,48 +53,44 @@ setMethod("eval_code", signature = c(object = "qenv.error"), function(object, co
     return(object)
   }
   code_split <- split_code(paste(code, collapse = "\n"))
+
   for (i in seq_along(code_split)) {
```
averissimo, line 57: I think we can evluate all the code split at once with this package.

It already gives the warnings / errors / outputs line by line.

We just need to map it after the fact, that is, outputs, errors, messages and warnings after source element are mapped to that source.

R/qenv-eval_code.R
```diff
@@ -53,48 +53,44 @@ setMethod("eval_code", signature = c(object = "qenv.error"), function(object, co
...
-        invokeRestart("muffleWarning")
-      },
-      message = function(m) {
-        attr(current_code, "message") <<- cli::ansi_strip(sprintf("> %s", conditionMessage(m)))
-        invokeRestart("muffleMessage")
-      }
+    x <- evaluate::evaluate(
+      current_code,
+      envir = object@.xData,
+      stop_on_error = 1,
```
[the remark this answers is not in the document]
gogonzo: Makes sense to run `evaluate::evaluate` on the whole code. There is one problem. `library(xyz)` won't have an effect for the next call as we manually do a following, to make sure that attached package will be over .xData. 

```r
  if (!identical(parent.env(object@.xData), parent.env(.GlobalEnv))) {
    # needed to make sure that @.xData is always a sibling of .GlobalEnv
    # could be changed when any new package is added to search path (through library or require call)
    parent.env(object@.xData) <- parent.env(.GlobalEnv)
  }
```

- Issue you pointed out can be avoided by calling all as a function call.
- Another (minor) issue when depending on the `evaluate::evaluate` code parser is that comments are treated independently and they are not associated with the particular code lines (we do this). For example:

```r
evaluate::evaluate('# overline comment \n this <- print("x") # inline comment')
# <evaluation>
# Source code: 
#   # over line comment 
# Source code: 
#    this <- print("x")
# Text output: 
#   [1] "x"

```

----------

In general I have an opinion that we should rather support grDevice (`new_device = TRUE`) and evaluate the whole code once in `evaluate::evaluate`. Problem with libraries we can explain easier in the documentation than problem of new device per call. I consider comments problem as minor, but I suspect that it might have some rare consequences with "# linksto ..."

R/qenv-eval_code.R
```diff
@@ -44,60 +44,55 @@ setMethod("eval_code", signature = c(object = "qenv.error"), function(object, co
   if (identical(trimws(code), "") || length(code) == 0) {
     return(object)
   }
+  code <- paste(split_code(code), collapse = "\n")
```
averissimo, line 47: Maybe we can simplify this with another `evaluate` function

```r
split_code <- function(code) {
  parsed_code <- evaluate::parse_all(code)
  unlist(
    mapply(
      function(src, expr) {
        pd <- utils::getParseData(parse(text = src, keep.source = TRUE))
        comments <- paste(trimws(pd[pd$token == "COMMENT", "text"]), collapse = " ")
        new_src <- trimws(paste(as.character(expr), comments))
      },
      parsed_code$src,
      parsed_code$expr,
      USE.NAMES = FALSE
    )
  )
}
```

This has the side effect of some code formatting, but we can avoid that by re-parsing the `src` column

averissimo: Not immediately, but it does as splits the expressions as 2 in the right column

My code above then adds the comment to all of them (we can keep just the last though)

``` r
split_code <- function(code) {
  parsed_code <- evaluate::parse_all(code)
  as.character(unlist(
    mapply(
      function(src, expr) {
        pd <- utils::getParseData(parse(text = src, keep.source = TRUE))
        comments <- paste(trimws(pd[pd$token == "COMMENT", "text"]), collapse = " ")
        new_src <- trimws(paste(as.character(expr), comments))
      },
      parsed_code$src,
      parsed_code$expr,
      USE.NAMES = FALSE
    ))
  )
}

code <- "y <- 0; j <- 'see quotes changing'; for (x in 2) { y <- y + x}"
split_code(code)
#> [1] "y <- 0"                            "j <- \"see quotes changing\""     
#> [3] "for (x in 2) {\n    y <- y + x\n}"
```

<sup>Created on 2025-06-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

NEWS.md
```diff
@@ -1,11 +1,16 @@
 # teal.code 0.6.1.9003
 
+### Enhancements
+
+* Introduced `get_outputs` function to fetch objects which have been printed or plotted in the `qenv` code.
```
averissimo, line 5: Should we add a small section on the vignette as well? in addition to "### Warnings and messages in `qenv` objects"

```diff
diff --git a/vignettes/qenv.Rmd b/vignettes/qenv.Rmd
index 4859e32c..302b6846 100644
--- a/vignettes/qenv.Rmd
+++ b/vignettes/qenv.Rmd
@@ -126,7 +126,7 @@ The feasibility of joining `qenv` objects hinges on the contents of the environm
 
 ### Warnings and messages in `qenv` objects
 
-In cases where warnings or messages arise while evaluating code within a `qenv` environment, these are captured and stored within the `qenv` object. Access these messages and warnings using below
+In cases where warnings or messages arise while evaluating code within a `qenv` environment, these are captured and stored within the `qenv` object. Access these messages and warnings using `get_messages()` and `get_warnings()` functions as shown below.
 
 ```{r}
 q_message <- eval_code(qenv(), quote(message("this is a message")))
@@ -138,6 +138,15 @@ get_warnings(q_warning)
 
 If any of above returns `NULL`m then no warnings nor messages were present.
 
+### Outputs in `qenv` objects
+
+When evaluating code in a `qenv`, the outputs are captured and stored in a list. This includes printed objects, plots, warnings, and messages. You can retrieve these outputs using the `get_outputs()` function.
+
+```{r}
+q_output <- eval_code(qenv(), quote(plot(1:10)))
+get_outputs(q_output)
+```
+
 ## Utilizing `qenv` inside `shiny` applications
 
 These functions can be seamlessly integrated into `shiny` applications to produce reproducible outputs. In the example below, the `rcode` section showcases the code employed for generating the output.
```

gogonzo: I've included some info about `get_outputs()`. Please have a look once again before I merge

PR #260, opened 2025-06-24
Move `library()` override back to `eval_code()`

averissimo:
@gogonzo This might be a better solution that keeps the override under `evaluate()`, while avoiding changing who the `parent.env` is.

That is, we don't need to check the "`grand.parent.env()`"


gogonzo:
Nice, thanks @averissimo . The API with this `include_funs` is ridiculous :D

PR #268, opened 2025-08-11
Update lifecycle badges

m7pr:
Both `experimental` and `deprecated` badges were added at the latest version 0.6.1 so no need for the updates https://github.com/insightsengineering/teal.code/pull/239
Only added `experimental` for the new function: `get_outputs`.


gogonzo:
> Both `experimental` and `deprecated` badges were added at the latest version 0.6.1 so no need for the updates #239 Only added `experimental` for the new function: `get_outputs`.

All deprecations have been added at 0.6.0 which means that we should `deprecate_stop` now

Skärmavbild 2025-08-11 kl  15 35 19

gogonzo:
> @gogonzo for the `deprecate_stop` do you think we should deleted examples, tests and body of the function (besides the lifecycle check)?

Yes

R/qenv-join.R
```diff
@@ -1,152 +1,23 @@
...
+#' @aliases join,qenv.error-method
 #'
 #' @export
-setGeneric("join", function(x, y) standardGeneric("join"))
-
-setMethod("join", signature = c("qenv", "qenv"), function(x, y) {
-  lifecycle::deprecate_soft("0.6.0", "join()", "c()")
-  c(x, y)
-})
+setGeneric("join", function(...) standardGeneric("join"))
```
gogonzo, line 15: same here, we don't need a method
```suggestion
join <- function(...) lifecycle::deprecate_stop("0.6.0", "join()", "c()")
```

R/qenv-get_var.R
```diff
@@ -5,45 +5,12 @@
...
-#' q <- qenv()
-#' q1 <- eval_code(q, code = quote(a <- 1))
-#' q2 <- eval_code(q1, code = "b <- a")
-#' get_var(q2, "b")
-#'
-#' @aliases get_var,qenv,character-method
-#' @aliases get_var,qenv.error,ANY-method
+#' @param ... function is deprecated.
+#' @param x (`qenv`)
+#' @param i (`character(1)`) variable name.
```
gogonzo, line 10:
```suggestion
```

gogonzo (approved):
👍

vedhav:
@insightsengineering/nest-core-dev TODO: Please update these places where `teal.code::get_var` are used/mentioned:
1. In teal we use it [here](https://github.com/insightsengineering/teal/blob/92466ce338839d68d5f622de19ccf9c45cca85a1/R/module_init_data.R#L100) and it is causing the app to fail.
2. In `teal.connectors.ssh` [a test](https://github.com/insightsengineering/teal.connectors.ssh/blob/d754f793a2c92b2d5efdcf2b787a73a74038adf9/tests/testthat/test-ssh_connector.R#L93) uses this.
3. `teal.data` docs mentioned it [here](https://github.com/insightsengineering/teal.data/blob/e78d3bf2fd168bd5ce63ee101bea8219a1b43868/R/teal_data-class.R#L19).

llrs-roche:
> @insightsengineering/nest-core-dev TODO: Please update these places where `teal.code::get_var` are used/mentioned:
> 
> 1. In teal we use it [here](https://github.com/insightsengineering/teal/blob/92466ce338839d68d5f622de19ccf9c45cca85a1/R/module_init_data.R#L100) and it is causing the app to fail. 

[Couldn't find on main](https://github.com/search?q=repo%3Ainsightsengineering%2Fteal+get_var&type=code) fixed on https://github.com/insightsengineering/teal/pull/1576

The other two aren't fixed. Maybe I'll create an issue to don't forget about those (specially because it has teal.connectors.ssh repo that I was not aware of). 

> 2. In `teal.connectors.ssh` [a test](https://github.com/insightsengineering/teal.connectors.ssh/blob/d754f793a2c92b2d5efdcf2b787a73a74038adf9/tests/testthat/test-ssh_connector.R#L93) uses this.
> 3. `teal.data` docs mentioned it [here](https://github.com/insightsengineering/teal.data/blob/e78d3bf2fd168bd5ce63ee101bea8219a1b43868/R/teal_data-class.R#L19).
Still there.

PR #274, opened 2025-10-17
Set expectation when no code is used with qenv

llrs-roche:
# Pull Request

Document expectation of what should return `qenv` with no code.
Discovered while working on https://github.com/insightsengineering/teal/pull/1630 where some code assumed it would return `NULL`. 

Previously it was only tested that `testthat::expect_identical(get_code(qenv(), names = "a"), "")`.
It makes sense that it also return `""` when there is no code.


osenan (approved):
This is a very useful test. Good job!

Issue #278, opened 2025-10-31
[Bug]: Functions with side effects are not detected as dependencies


llrs-roche:
### What happened?

Functions with side effects like `options` and `set.seed` are omitted on teal Show R code but they are needed to make the code reproducible:

```r
data <- within(teal.data::teal_data(), { 
  ADSL <- teal.data::rADSL
  set.seed(1000)
  options(test = 1)
  ind <- sample(nrow(ADSL), 5)
  ADSL$STUDYID[ind] <- 1L
})

app <- teal::init(data = data, modules = teal::example_module())
runApp(app)
```

The code on apps will omit the second and third lines. But when user copies and pastes the code result will be different as it will use a different seed and the check step ( `stopifnot(rlang::hash(ADSL) == "...`) will fail. 

`srv_source_code()` doesn't return those lines because on teal app because on the reactive `module_teal_data` of the `module()` function any code that is not a dependency of the data required is dropped (L468 of module_nested_tabs.R). 

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

llrs-roche:
We have replaced on code and examples to `within`, pointing this now might cause some problems.
We might still get the comment with `utils::getParseData` if we set `option(keep.source = TRUE)` inside within. 
However, I agree that more documentation around this feature/corner case would be helpful. It is only mentioned in a line. 

BTW if we improve documentation we should add the `within` alias for this method. This way users will find it with `?within`.

PR #280, opened 2025-12-05
Expand documentation for the use and reproducibility of side effects functions in qenv

osenan:
# Pull Request

It addresses this issue #278 

While the reproducibility of `within` function is yet not supported with side effects functions, this PR will help users to make sure their code is reproducible or receive warning instead. In further issues we can debate whether we refactor within to support linking of side effects.

To test, verify that the vignette is correctly updated, see if tests are passed.

If you create a qenv with this branch with this code, it should trigger a warning:

```
q <- qenv() |> within({
  set.seed(2)
  a <- runif(1)
  b <- runif(1)
})
```


R/qenv-within.R
```diff
@@ -48,8 +55,10 @@
 #'
 within.qenv <- function(data, expr, ...) {
   expr <- as.expression(substitute(expr))
+  if (grepl(paste(side_effect_functions, collapse = "|"), as.character(expr))) {
+    warning("Calls to options and set.seed might not be reproducible with `get_code`, use `eval_code` instead")
+  }
```
llrs-roche: How does this new warning affect all the examples of teal.gallery and teal.modules.* packages?

m7pr:
@osenan did you see this : p? https://github.com/insightsengineering/teal.code/pull/280#issuecomment-3617267773

PR #281, opened 2025-12-11
Add copilot-setup-steps workflow

m7pr:
Add copilot-setup-steps workflow using the reusable template from r.pkg.template.


R/utils-get_code_dependency.R
```diff
@@ -604,4 +604,4 @@ split_code <- function(code) {
   # line split happens before call terminator (it could be `;` or `\n`) and the terminator goes to the next line
   # we need to move remove leading and add \n instead when combining calls
   c(new_code[1], gsub("^[\t ]*(\n|;)", "", new_code[-1]))
-}
+}
```
m7pr:
```suggestion
}

```

PR #286, opened 2026-04-23
comma added to q2

shiaoligreen:
# Pull Request

Fixes #247

PR #287, opened 2026-04-23
rerun docs and update news

donyunardi:
I was too fast when approving #286 

In this PR, I'm updating docs and add records in news. 
Not sure why CI/CD passing R CMD check (which need to look into it).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 ### Bug fixes
 
 * Fixed a problem parsing Chinese characters due to the encoding (#284)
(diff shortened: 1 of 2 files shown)


osenan (approved):
LGTM!
For the investigation on why the CI did not fail, running `devtools::check()` on `main` before this PR does not fail locally. This suggests that there is nothing wrong in the CI configuration, but the check itself does not fail in such small difference? It might be that the check does not fail for a single comma? But we should confirm it.

PR #294, opened 2026-09-07
Fix bug, add tests and update NEWS

### LINKED ISSUE #293: [Bug]: `[` in assignments causing missing code in show R code
### What happened?

This is not picking up the dependency properly (it's the `[`)

```r
library(teal.code)

data <- teal_data()
data <- within(data, {
  mtcars <- mtcars
  mtcars <- mtcars[mtcars$cyl == 4, ]
})
get_code(data, names = "mtcars")
# Expected: both lines
# Actual:   "mtcars <- mtcars" only — filter line dropped
```

The equivalent for dplyr::filter works

```r
data2 <- teal_data()
data2 <- within(data2, {
  mtcars <- mtcars
  mtcars <- dplyr::filter(mtcars, cyl == 4)
})
get_code(data2, names = "mtcars")
# Returns both lines correctly
```

Image

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

llrs-roche:
# Pull Request

Fixes #293

This is a minimal patch (mostly generated with AI) that fixes the test cases by skipping objects that are on the LHS.

This has revealed several issues that I'm considering what to do with it:
-  `extract_dependency()` list objects not present (cyl column on the first test added).
- duplicated parsing of code on `extract_dependency()`

### DIFF
--- a/tests/testthat/test-qenv_get_code.R
+++ b/tests/testthat/test-qenv_get_code.R
@@ -1,17 +1,17 @@
 pasten <<- function(...) paste(..., collapse = "\n")
 
-testthat::test_that("get_code returns character of length 0 if no code", {
+test_that("get_code returns character of length 0 if no code", {
   expect_identical(get_code(qenv()), "")
 })
 
-testthat::test_that("get_code returns code (character(1) by default) of qenv object", {
+test_that("get_code returns code (character(1) by default) of qenv object", {
   q <- qenv()
   q <- eval_code(q, quote(x <- 1))
   q <- eval_code(q, quote(y <- x))
-  testthat::expect_equal(get_code(q), pasten(c("x <- 1", "y <- x")))
+  expect_equal(get_code(q), pasten(c("x <- 1", "y <- x")))
 })
 
-testthat::test_that("get_code handles code elements being code-blocks", {
+test_that("get_code handles code elements being code-blocks", {
   q <- qenv()
   q <- eval_code(q, quote(x <- 1))
   q <- eval_code(
@@ -21,20 +21,20 @@ testthat::test_that("get_code handles code elements being code-blocks", {
       z <- 5
     })
   )
-  testthat::expect_equal(get_code(q), pasten(c("x <- 1", "y <- x", "z <- 5")))
+  expect_equal(get_code(q), pasten(c("x <- 1", "y <- x", "z <- 5")))
 })
 
-testthat::test_that("get_code returns expression of qenv object if deparse = FALSE", {
+test_that("get_code returns expression of qenv object if deparse = FALSE", {
   q <- qenv()
   q <- eval_code(q, quote(x <- 1))
   q <- eval_code(q, quote(y <- x))
-  testthat::expect_equivalent(
+  expect_equivalent(
     toString(get_code(q, deparse = FALSE)),
     "{\n    x <- 1\n    y <- x\n}"
   )
 })
 
-testthat::test_that("get_code called with qenv.error returns error with trace in error message", {
+test_that("get_code called with qenv.error returns error with trace in error message", {
   q1 <- qenv()
   q1 <- eval_code(q1, quote(x <- 1))
   q2 <- eval_code(q1, quote(y <- x))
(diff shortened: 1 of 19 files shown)


averissimo (commented):
This block is not very good, it works, but barely. It doesn't make sense to replace first instance after assignment, it would make sense to replace first instance in parenthesis.

See example below that incorrectly replaces first occurrence (the one not in parenthesis)

`browser()` statement was placed just before returned modified `result`

Image

```r
pkgload::load_all(".")

data <- qenv()
data <- within(data, {
  mtcars <- datasets::mtcars
  #mtcars <- mtcars[mtcars$cyl == 4, ]
  aa <- seq_len(nrow(mtcars))
  mtcars <- mtcars$mpg > 15 & aa[mtcars$mpg & mtcars$cyl]
  # mtcars <- mtcars[mtcars$cyl == 4, ]
})
get_code(data, names = "mtcars") |> cat()
```

Could we simplify the code to just return `c(ans, in_parenthesis(pd))`

We don't really care about having repeated element, or do we?

```diff
diff --git a/R/utils-get_code_dependency.R b/R/utils-get_code_dependency.R
index 1bd38ef0..5b2ac04d 100644
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -304,24 +304,7 @@ extract_occurrence <- function(pd) {
   after <- match(min(x$id[assign_cond]), sort(x$id[c(min(assign_cond), sym_cond)])) - 1
   ans <- append(x[sym_cond, "text"], "<-", after = max(1, after))
   ans <- move_functions_after_arrow(ans, unique(x[sym_fc_cond, "text"]))
-  roll <- in_parenthesis(pd)
-  if (length(roll)) {
-    # detect elements appeared in parenthesis and move them on RHS
-    # but only their first appearance after assignment
-    # as the same object can appear as regular object and the one used in parenthesis
-    result <- ans
-    arrow_idx <- which(result == "<-")
-    for (elem in roll) {
-      idx <- which(result == elem)
-      idx <- idx[idx > arrow_idx][1]
-      if (!is.na(idx)) {
-        result <- result[-idx]
-      }
-    }
-    c(result, roll)
-  } else {
-    ans
-  }
+  c(ans, in_parenthesis(pd))
 }
 
 #' Moves function names to the right side of dependency graph
```

Image

llrs-roche:
Great simplification! I think we do care about the position of the elements and using `unique()` would alter that, I only used it for the `in_parenthesis`. I like your approach much more than mine, it removes code without a side effect.

averissimo (approved):
💯

tests/testthat/test-utils-get_code_dependency.R
```diff
@@ -245,3 +245,23 @@ testthat::describe("get_code with subassignments", {
     testthat::expect_equal(get_code(td, names = "vec"), code_source)
   })
 })
+
+
+describe("get_code with subsetting", {
+  it("same object", {
```
averissimo, line 251: Can you add an extra case with the example I provided. It won't add that much, but it might uncover regressions in the future

tests/testthat/test-utils-get_code_dependency.R
```diff
@@ -245,3 +245,23 @@ testthat::describe("get_code with subassignments", {
     testthat::expect_equal(get_code(td, names = "vec"), code_source)
   })
 })
+
+
+describe("get_code with subsetting", {
```
averissimo, line 250:
```suggestion
testthat::describe("get_code with subsetting", {
```

averissimo: To keep in line with the rest of the package

llrs-roche: Mmh, we want to have the prefix? I thought we decided to go without prefixes on the framework. I can do it either way

tests/testthat/test-utils-get_code_dependency.R
llrs-roche: Added with 87f47c38

tests/testthat/test-utils-get_code_dependency.R
averissimo: I believe the decision was to prefer no prefixes, but be consistent on the test file, this one was has prefixes :shrug:

llrs-roche: I removed from the whole testthat folder at ed6463fd83490d4997f725ba4e44f0e88e6be029

tests/testthat/test-qenv_get_code.R
```diff
@@ -1,17 +1,17 @@
 pasten <<- function(...) paste(..., collapse = "\n")
```
averissimo: Can we fix this warning?

It passes locally without `<<-`

```suggestion
pasten <- function(...) paste(..., collapse = "\n")
```

averissimo: Check this branch: `lint_styler@293_subsetting`. It removes generic lintr exceptions and styler

https://github.com/insightsengineering/teal.code/compare/293_subsetting...lint-styler%40293_subsetting?expand=1

llrs-roche: I fixed the double assignment. But I won't fix the lintr issues here. It is strange, when I remove locally the .lintr and run it I see several `line_length_linter` (80 characters) and some `object_usage_linter` but no `semicolon_linter`, `assigment`or `brace_linter`.

